PR #2, opened 2015-05-08
Fixes broken links in the README

cameronviner:
Seems the links in the read me have become broken when this was moved over, this just points them to the correct place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,16 +5,16 @@
 This is a collection of documents that outline Holiday Extras' culture and developer expectations.
 
 ## Contents
- * [General Javascript Best Practices](https://github.com/holidayextras/hx-culture/blob/master/general-javascript-best-practices.md)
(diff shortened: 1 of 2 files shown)


lukefrake:
Good shout, my mistake.

I've made the same changes in another file and pushed into here as well.

hxpaul:
could these be made relative links?

cameronviner:
@hxpaul Yep have just made that change could you have another look when you get a second

hxpaul:
:+1:

pmcnr-hx:
I would remove the redundant preceding `/` from the file names in the links for these to actually be relative to `README.md`.

hpoom:
Merged as people are commenting on broken links.

@pmcnr-hx If you think it still need changing do another PR

pmcnr-hx:
Agreed @hpoom: it's merely cosmetic. It can always be improved in the context of a future PR.

PR #4, opened 2015-05-13
Add PR best practices.

michaelcarter:
Nothing to say, it's all in the files changed. :arrow_down:

### DIFF
--- a/pr-best-practices.md
+++ b/pr-best-practices.md
@@ -1,0 +1,11 @@
+# PR Best Practices
+
+## Keep them small
+
+Keep your pull requests as small as possible. We should be looking for ways to split up our work into the smallest units that are sensible to merge. Some things we should be thinking:
+
+- *"Are there any small code refactors I can do as a preliminary PR into edge that will make completing this feature a trivial task?"*
+- *"Are there sub-features of this Jira ticket that can be deployed in isolation and have benefit to the product before the rest of my work is done?"*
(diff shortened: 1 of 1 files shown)


pr-best-practices.md
```diff
@@ -0,0 +1,11 @@
+# PR Best Practices
+
+## Keep them small
+
+Keep your pull requests as small as possible. We should be looking for ways to split up our work into the smallest units that are sensible to merge. Some things we should be thinking:
+
+- *"Are there any small code refactors I can do as a preliminary PR into edge that will make completing this feature a trivial task?"*
+- *"Are there sub-features of this Jira ticket that can be deployed in isolation and have benefit to the product before the rest of my work is done?"*
+- *"Perhaps I should merge lots of small pull requests into a separate branch for my feature and have those reviewed individually, and then merge that into edge once it's tested."*
```
hx-markterry, line 9: mega-de and heha got hit by this approach recently. You do still end up with a big code change in a single deploy.

michaelcarter, line 9: Yes, but this in combination with other approaches outlined here might have been a better option for mega-de. In hindsight it looks like there were opportunities there for deploying things as separate smaller refactors or features in support of the overall goal.

pmcnr-hx, line 9: I don't think there is a _silver bullet_ solution for the problem of big features development. We're developing the OMS in a separate branch because it's pretty much self-contained, but I would feel uncomfortable merging it to master when it is ready if it weren't for that. _Integrate early and often_ is best when it is possible.

PR #5, opened 2015-05-13
Add guidance to "contributing" section

shackpank:
# What does this change?

Adds a very simple PR template and a note on when to consider changes mergeable to the contributing section of the README.
# Why?

Recent PRs have used the PR template with all its checkboxes about unit tests and cross-browser checks when adding documentation here, we'd like to make it clear that's not necessary. Similarly, the 3-approvals process with review slots that need to be taken and potentially chased up is less suited to documentation updates than it is to code changes.


hpoom:
Love this addition gets a :+1: from me

stuartstobie:
This makes a good deal of sense, gets a :+1: from me.

PR #7, opened 2015-05-19
guidelines & gotchas for converting coffee to js

jackdclark:
# What does this change?

Introduces some guidelines for converting CoffeeScript to JavaScript, to help with the imminent refactor of the trip app (and potentially cordova app)
# Why?
- Less of a learning curve, _everybody knows JavaScript_
- Faster compilation times (eventually)
- Easier to convert from requirejs to browserify


hx-markterry:
I think there should be a reference to our other js guidelines from this doc, to remind devs want not to do

coffeescript-to-javascript-guidelines.md
```diff
@@ -0,0 +1,168 @@
...
+// generated by coffee-script 1.9.2
+```
+
+#### Refactored JavaScript
+This is fairly unreadable, we would remove the temporary `ref` variables and do the following
+
+```javascript
+_hasFlightCode: function() {
+  var flights = this.get('flights');
+  if(!(flights && flights.out && flights.out.flight && flights.out.flight.code)) {
```
dancork: This sort of goes against our [culture docs](https://github.com/holidayextras/culture/blob/master/serverside-javascript-best-practices.md#keep-the-logic-simple) of keeping it simple.

I know that's the server side doc but I feel it should apply, something similar to this... 

```
if(!flights) {
  return;
}
if(!flights.out) {
  return;
}
if(!flights.out.flight) {
  return;
}
return flights.out.flight.code;
```

coffeescript-to-javascript-guidelines.md
```diff
@@ -0,0 +1,168 @@
...
+  };
+})(this)), 100);
+```
+
+#### Refactored JavaScript
+Rather than wrapping & passing `this` into the scope of the callback, we would assign `this` to `self` outside the scope of `debounce`, like so
+
+```javascript
+var self = this;
+_.debounce(function() {
```
dancork: personally i'd favour scoping it still. can be done using underscore/lodash bind.

```
_.debounce(
  _.bind(function(){
    return self.trigger('scroll');
  }, this)
, 10);
```

coffeescript-to-javascript-guidelines.md
jackdclark: This is probably personal preference but I find the original easier to follow? It tends to be the style I've seen when working on our JS projects. This is probably just an extreme example of a deeply nested property, hard to get around :(.

coffeescript-to-javascript-guidelines.md
jackdclark: Either or I'd say, I've favoured the `self` implementation before as I've heard of performance problems with the native `bind` functionality, don't know if the lodash one has the same though? Happy to review this, or add both as an example?

dancork: The reason I'd favour bind would be for consistency across the app. It allows for anonymous or named functions.

```
_.debounce(
  _.bind(function(){
    return self.trigger('scroll');
  }, this)
, 10);
```

could easily be refactored to...

```
_.debounce( _.bind(_myScrollyFunction, this), 10);
```

coffeescript-to-javascript-guidelines.md
dancork: True, there is a fair amount of subjectivity to this. I'm not a big fan of the single exclamation for negative comparison of multiple objects.

Not to harp on, but the culture repo does state that we should strive for [https://github.com/holidayextras/culture/blob/master/coding-principles.md#clarity-over-brevity](clarity over brevity) which I don't think we have in this example. 

`if (!flights && !flights.out && !flights.out.flight && !flights.out.flight.code) {` is probably a more readable solution.

michaelcarter:
We should lean on our culture docs for what good code looks like as much as possible. If there's disagreement there then that should be taken up as a PR into the culture repo, but PRs should reflect the coding standards we have laid out - it's our source of truth and keeps these discussions from increasing time to production.

simpsons

coffeescript-to-javascript-guidelines.md
pmcnr-hx: Actually @dancork:

```
!(flights && flights.out && flights.out.flight && flights.out.flight.code)
```

is logically equivalent to:

```
!flights || !flights.out || !flights.out.flight || !flights.out.flight.code
```

jackdclark:
@dancork @michaelcarter @hx-markterry  I've added a note at the top of the conversion section that links to our coding principles & JavaScript guidelines so that this file doesn't get treated as a style guide itself. More low level discussions around exactly what to use (e.g. `bind` vs `var self = this`) should happen there, this file is only intended as a tool to help pick out the _magic_ that CoffeeScript generates.

Are you able to give this another look?

hx-markterry:
ok with me :guitar:

PR #9, opened 2015-05-26
Adding a contributing template

jodiedoubleday:
# What does this change?

Adds a new contributing-template.md for us to use with our open source projects.
# Why?

We should have documents and guidelines surrounding our open source repos, at the moment we don't so here are some. We should include this in ALL our open source projects.


cameronviner:
looks good :+1:

andycastleuk:
Looks good :+1:

PR #15, opened 2015-06-08
Dealing with coffeescript return values

hx-markterry:
Coffeescript always returns values from functions, but in javascript it might not be required, this PR makes a point about this and asks the developer to check if the return statements are really required.
- [x] Someone is ok with this.
- [x] Another human this is is a good idea.


jackdclark:
Yes sir :+1:

hxpaul:
:+1:

PR #17, opened 2015-06-17
Abba Split Testing

dancork:
I've recently made a change to Abba in the tripapp. Here is a doc to support that.

### DIFF
--- a/split-test-policy.md
+++ b/split-test-policy.md
@@ -35,7 +35,7 @@ abba.test('My test').start();
 Though if you require customisation you can set these yourself.
 
 ```javascript
-// start a simple ab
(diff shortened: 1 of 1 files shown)


SkylerAvery:
:+1:

lukefrake:
Can you call this split test policy? Or really should it be how to use Abba?
Dune I read split test policy I thought it would be rules about how we split test, what we split test etc.

If this is a how to, shouldn't it live in the Abba repo instead of here?

dancork:
Good point @lukefrake, will move it to tripapp repo for now. In future plan is to extract it and make it an npm module but until then should live there.

Cheers.

PR #18, opened 2015-06-23
Proposal for serverside best practices + promises

hxoliverrumbelow:
A proposal for allowing promises within our best practices doc without compromising too much on quality.

### DIFF
--- a/serverside-javascript-best-practices.md
+++ b/serverside-javascript-best-practices.md
@@ -91,6 +91,36 @@ In the above example its easy to identify a module's dependencies and the module
 
 
 
+## Promises or Callbacks?
+
+Provided the orchestration of asynchronous actions are decoupled from the code that provides functionality, both conventions are acceptable. For example, both of the following examples are great:
+
+```javascript
+ourModuleName._orchestrate = function() {
+  return ourModuleName._firstThing()
+    .then(ourModuleName._secondThing)
+    .then(ourModuleName._thirdThing)
+    .then(ourModuleName._fourthThing)
+    .fail(function(err) {
+      //
+    });
+};
+```
+```javascript
+ourModuleName._orchestrate = function(callback) {
+  async.waterfall([
+    ourModuleName._firstThing,
+    ourModuleName._secondThing,
+    ourModuleName._thirdThing,
+    ourModuleName._fourthThing
+  ], callback);
+};
+```
+
+
+
+
+
 ## Return Early
 
 * Each logic block is isolated
@@ -255,7 +285,7 @@ obj._request = function(p1, p2, callback) {
 
 
 
-## Handling heavily sequential async code
+## Handling heavily sequential async code with callbacks
 
 * `async.waterfall` allows us to chain functions together
 * The closures around `_task1/2/3/4` allow us to pass data over each private function instead of needing to proxy data through functions.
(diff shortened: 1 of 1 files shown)


serverside-javascript-best-practices.md
```diff
@@ -320,6 +350,53 @@ createBooking.call(self, product, params, 'upgrades', 'supplier', 'payment', fun
...
+createBooking._orchestrate = function(data) {
+  return createBooking._sendBook1Request(data)
+    .then(createBooking._generatePaymentRequest)
+    .then(createBooking._processPayment)
+    .then(createBooking._sendBook2Request)
+    .then(createBooking._generateBookingReference);
+};
+```
+```javascript
+require('sinon-bluebird');
```
pmcnr-hx, line 347: [`sinon-as-promised`](https://www.npmjs.com/package/sinon-as-promised) might be the best option here since it works with different promises implementations, as long the implementation provides a constructor, which should cover ES6 promises and the most popular promises libraries (Bluebird, Q, ...).
On the other hand, if Bluebird is the only promises library used, `sinon-bluebird` provides more advanced features, so maybe making clear that the given example is using Bluebird instead and keeping `sinon-bluebird` may work equally well.

serverside-javascript-best-practices.md
```diff
@@ -320,6 +350,53 @@ createBooking.call(self, product, params, 'upgrades', 'supplier', 'payment', fun
 
 
 
+## Handling heavily sequential async code with promises
+
+* !! Using promises for heavily sequential async code has the downside of hiding the data being passed between functions. We lose the ability to wrap each layer in a closure to selectively pass the correct data betweeen the correct entites. Eveything must flow through the middle to get to the end. This breaks our rules on single-responsibility, but allows for slightly cleaner orchestration:
```
pmcnr-hx: What is the purpose of the `!!` prefix in this list's items?

pmcnr-hx: `betweeen` → `between`

pmcnr-hx: `Eveything` → `Everything`

serverside-javascript-best-practices.md
```diff
@@ -320,6 +350,53 @@ createBooking.call(self, product, params, 'upgrades', 'supplier', 'payment', fun
...
+* !! Sinon promises do not support sandboxing yet, so you'll have to manually restore everything by hand
+* !! Its impossible to prove that the data flow between promises is correct, only that data moves between them.
```
pmcnr-hx: `Its` → `It's`

serverside-javascript-best-practices.md
```diff
@@ -320,6 +350,53 @@ createBooking.call(self, product, params, 'upgrades', 'supplier', 'payment', fun
...
+```
+```javascript
+require('sinon-bluebird');
+
+sinon.stub(createBooking, '_sendBook1Request').resolves('data1');
+sinon.stub(createBooking, '_generatePaymentRequest').resolves('data2');
+sinon.stub(createBooking, '_processPayment').resolves('data3');
+sinon.stub(createBooking, '_sendBook2Request').resolves('data4');
+sinon.stub(createBooking, '_generateBookingReference').resolves('data5');
+
```
kevinhodges, line 354: Looks tidy in a basic use case but once you start modelling a more real world example things will get messy pretty quickly.  Stubbing objects and multiple arguments for more than one function at a time is gonna get out of hand, not sure that's avoidable though.  I'd suggest using `.spread` instead of `.then` for this kind of pattern, gives greater visibility on what's being passed between the methods.

serverside-javascript-best-practices.md
hxoliverrumbelow, line 347: I'm keen to retain the `.calledWithPromise( )` feature, or rather, I'm happy with any solution that provides the ability to assert what a stub was invoked with. For now putting in a few lines to say this example uses Bluebird to achieve that feels like the best approach.

pmcnr-hx:
Looks good to me: :+1:

rahulpatel:
I like it! :+1:

serverside-javascript-best-practices.md
```diff
@@ -320,6 +350,54 @@ createBooking.call(self, product, params, 'upgrades', 'supplier', 'payment', fun
 
 
 
+## Handling heavily sequential async code with promises
+
+* !! Using promises for heavily sequential async code has the downside of hiding the data being passed between functions. We lose the ability to wrap each layer in a closure to selectively pass the correct data between the correct entites. Everything must flow through the middle to get to the end. This breaks our rules on single-responsibility, but allows for slightly cleaner orchestration:
+* !! Sinon promises do not support sandboxing yet, so you'll have to manually restore everything by hand
+* !! It's impossible to prove that the data flow between promises is correct, only that data moves between them.
```
michaelcarter: I think some code or even links to articles demonstrating these issues would be useful. Particularly, where you've said `"it's impossible"`, `"we lose the ability"` and `"everything must"` - I think it's important developers understand. 

For example I'm not sure why it's _"impossible to prove the flow of data between promises is correct"_. I'm wondering why we can't we just spy on them and see what they're called with and what they resolve or don't resolve with?

michaelcarter: I've chatted to @hxoliverrumbelow about this and we've talked about including examples in an appendix at the bottom of this document so the message isn't cluttered up with additional information. :+1:

michaelcarter:
Looks good :+1: Appreciate the detailed explanation in the Appendix.

PR #19, opened 2015-06-29
More linting

hx-markterry:

Provides a more comprehensive ruleset, rather than the many little steps approach.

None
#### How can be this tested?

Manually:

```
npm install eslint
eslint -c ~/git/culture/.eslintrc test.js
```

CI:
Using the `make-up` util is the preferred way, this will be explained once we have a rule set we can agree on. 

---
#### Developer Definition of Done/Quality Checklist (for PR author to complete BEFORE code review):

Please only put your tick below if you are 100% behind these rules...
- [x] Ollie B
- [x] Oliver R
- [x] Mike
- [x] Kevin
- [ ] Rob J
- [x] Jodie

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -28,6 +28,13 @@
     "comma-style": [2, "last"],
     "indent": [2, 2],
     "quotes": [2, "single", "avoid-escape"],
-    "linebreak-style": [2, "unix"]
+    "linebreak-style": [2, "unix"],
+    "brace-style": [2, "1tbs"],
+    "no-new-object": 2,
+    "no-nested-ternary": 2,
+    "no-unneeded-ternary": 2,
+    "object-curly-spacing": [2, "always"],
+    "space-after-keywords": [2, "always"],
+    "space-before-function-paren": [2, "never"]
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,7 +18,6 @@
     "no-underscore-dangle": 0,
     "curly": [2, "multi-line"],
     "camelcase": [2, {"properties": "never"}],
-    "space-in-parens": [2, "always"],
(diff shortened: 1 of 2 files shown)


kevinhodges:
Your contributing document link is broken.

hx-markterry:
Yup sorry about that, i left in a part of the PR template we use for other repos, will have to look into having a contributing doc for this repo too

.eslintrc
```diff
@@ -15,17 +15,27 @@
     "jsx": true
   },
   "rules": {
-    "quotes": 0,
     "no-underscore-dangle": 0,
     "curly": [2, "multi-line"],
     "camelcase": [2, {"properties": "never"}],
-    "space-in-parens": [2, "never"],
+    "space-in-parens": [2, "always"],
```
shackpank: This is the only one I see as controversial - I think we're split distribution/SB on it (using "never" in distribution code). I'm of the opinion the extra spacing adds noise rather than improves readability, others the opposite, and it's easy to find styleguides from prominent organisations recommending either option.

hxoliverrumbelow:
:+1:

.eslintrc
michaelcarter: I think we're putting this in to appease short breaks, who use this in their current linting rules. Nobody over at the wave was so concerned about it that they 100% wanted it kept out.

jodiedoubleday: I would argue that neither team have come up with a reason why either is good or important. Therefore i think we should potentially leave this out of linting rules.

michaelcarter:
+1

jodiedoubleday:
other than above comment :+1:

hx-markterry:
I've now removed the controversial rule, so with or without spacing will be allowed. I'm more concerned with having a single agreed ruleset

PR #20, opened 2015-07-16
Drop consistent-return eslint rule

hxoliverrumbelow:
The eslint rule `consistent-return` defaults to enabled. In short, if a function does a `return` at any point, it requires all code paths to `return`. Here's their example: 

```
function doSomething(condition) {
    if (condition) {
        return true;
    } else {
        return;
    }
}
```

I'd like to drop this default rule for 4 reasons:
- Within async functions (for reference 38% of function declarations in HAPI are async) we're using the `return` statement to prevent further code execution, not to return any data. eslint can't tell that `callback`s never return anything, so it's static analysis complains of the missing `returns` when we invoke an async function.
- Our best practices doc clearly states that functions must either `return` data, or `callback` with data. We will not accept code that combines both: https://github.com/holidayextras/culture/blob/master/serverside-javascript-best-practices.md#functions-must-either-return-data-or-callback-with-data
- We expect test coverage over each return statement, all return paths of synchronous functions will have assertions to prove the correct data is being return'd: https://github.com/holidayextras/culture/blob/master/testing-standard.md#each-test-should-1
- When using an async function, we have to plaster additional `return`s before every single async function invocation. This makes our code harder to read and does not add any value. Here's a very simple example for reference: 

```
// before
var doSomeStuff = function(params, callback) {
  if (!params.something) return callback("Invalid request");

  otherModule.get(params.something, function(err, newItems) {
    if (err) return callback(err);
    var flatArray = [].concat.apply([], newItems);

    return callback(null, flatArray);
  });
};

// after
var doSomeStuff = function(params, callback) {
  if (!params.something) return callback("Invalid request");

  return otherModule.get(params.something, function(err, newItems) {
    if (err) return callback(err);
    var flatArray = [].concat.apply([], newItems);

    return callback(null, flatArray);
  });
};
```
- [ ] MarkT
- [ ] MikeC
- [x] OllieB
- [ ] Jodie
- [ ] Kev

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,6 +35,7 @@
     "object-curly-spacing": [2, "always"],
     "space-after-keywords": [2, "always"],
     "space-before-function-paren": [2, "never"],
-    "valid-jsdoc": 2
+    "valid-jsdoc": 2,
+    "consistent-return": 0
   }
 }


shackpank:
All good reasons :+1:

hpoom:
:+1:

hxoliverrumbelow:
This might be a red herring, Pedro and I are hopping versions of eslint and other environment-based linters / UI plugins. I'm going to close this for now and re-open if it comes back. Right now it looks like something else was causing the lint errors. Thanks to @kevinhodges for running the example through his own linting :+1:

PR #21, opened 2015-07-24
Add `semi` rule to `.eslintrc`

pmcnr-hx:
According to the ESLint documentation the [`semi` (_Enforce Semicolons_)](http://eslint.org/docs/rules/semi.html) rule should be enabled by default, but from my usage experience it seems to need to be enabled explicitly.

This pull request reorders the existing rules automatically for ease of maintenance and enables the `semi` rule explicitly, which I believe it is an important rule to enforce, since missing semicolons can lead to erroneous behaviour.


lukefrake:
:+1:

shackpank:
+1

PR #22, opened 2015-07-28
Include AMD-to-commonjs in coffee to js guidelines

shackpank:
We've made the transition from requirejs to browserify on a large internal project using a transform, which let us land it in master without changing enormous numbers of files but leaves us with a mixture of module styles.

Converting from one module style to the other is a simple enough process that we can opportunistically hook it on the ongoing coffee-to-js work and start making some headway on reaching a point where the application is all commonjs & we don't need the browserify transform anymore.

I've successfully tried out the commonjs conversion on an open coffee-to-js pull request, which is the starting point for these guidelines.

### DIFF
--- a/coffeescript-to-javascript-guidelines.md
+++ b/coffeescript-to-javascript-guidelines.md
@@ -2,11 +2,45 @@
 
 ## Steps
 1. Rename your file extension from `foo.coffee` to `foo.js`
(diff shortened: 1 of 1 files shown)


Joezo:
:+1:

jackdclark:
:+1:

PR #23, opened 2015-07-30
static method guidelines for :coffee: conversion

jackdclark:
Came across some static methods when converting our bookings collections in the trip app project. Added some guidelines for converting this.

> There are occasions when you will come across static methods or properties that are attached directly to the class itself. We can use the optional `classProperties` parameter in [Backbone's extend function](http://backbonejs.org/#Collection-extend) to achieve the same thing.


shackpank:
+1

rahulpatel:
Nice one! +1

DuncanFenning:
Fine by me! :+1:

PR #27, opened 2015-08-14
Specify all current default rules

hx-markterry:

We are currently using version 0.x of eslint which turns a lot of rules on automatically. This feature has been dropped as of 1.0. 

In order increase visibility of the enforced rules and prepare for a move to 1.0 of eslint I've now defined the default rules in the file.

See this link for reference:

http://eslint.org/docs/user-guide/migrating-to-1.0.0

N/A
#### How can this be tested?

The old and new rulesets can be compared in the following way:
1. Download the ruleset in master as `/tmp/master_rules`
2. Download the ruleset from this branch as `/tmp/branch_rules`
3. Checkout the `make-up` repo and change into it
4. Run `script/calculate_eslint_config.js /tmp/master_rules > /tmp/old`
5. Run `script/calculate_eslint_config.js /tmp/branch_rules > /tmp/new`
6. Run `diff /tmp/old /tmp/new`

giphy

---
- [x] 1 +1
- [x] 2 +1


pmcnr-hx:
This is definitely a good thing! My recent pull request was precisely because I hadn't really noted the whole new _all rules off by default_ in 1.x (which btw, with new _extends_ functionally still allowing the easy use of a default ruleset is the better behaviour since it removes the magic of not knowing clearly which rules are enabled or not) combined with the fact that make-up ruleset was still for ESLint < 1.x. :+1:

shackpank:
Works for me :+1:

PR #28, opened 2015-08-14
Added general meeting to the repo

robhuzzey:
I wanted to add something on here to help capture the culture from some of the great meetings I have attended.

This isn't fully formed, it's a start.


pmcnr-hx:
I agree with the usefulness of this document and think the current content is a good start! Looking forward to more comments.

jodiedoubleday:
I think we should have something on there for time-boxing the meeting and ensuring we have a chair so the meeting doesn't get out of hand.

robhuzzey:
@jodiedoubleday thanks... I've added a few more points, see what you think.

hx-markterry:
This looks good, and has no more objections, @robhuzzey could you merge it please.

PR #33, opened 2015-08-25
The new blameless postmortem...

alannajenkins:
After running a blameless postmortem last week, I wanted to take the opportunity to add this to the culture repo for easier access and update some of the language, as it felt a little harsh given what we are trying to achieve. 

I've also condensed some of the points of discussion as the format felt a little repetitive in places. Hopefully everyone feels we'll still get the same information out of the headers that I've changed to. 

Wanted to get everyone's opinion, rather than just putting it straight into the repo.


Incident Wash-Up
```diff
@@ -0,0 +1,19 @@
+Incident Wash-Ups should be used to discuss issues that have affected our customers and how we can help to avoid similar situations from occurring in the future. The Wash-Up should include all relevant parties at their earliest opportunity after the event. 
+
+Please use this format for all Incident Wash-Ups so that we have a consistent set of data on how we handle situations. Please add all notes from Incident Wash-Ups to the Culture repo, within the Wiki, for reference.
```
hx-markterry: the culture repo is public do we really want to add all the notes there?

jodiedoubleday: agree, we wouldn't keep notes of outcomes int eh culture repo, maybe in project repos or in its own space

SkylerAvery: Why not make a confluence page for all the incident washups?

kevinhodges: In SB we've created a type of jira so we can easily link any pieces of work that drop out or people that were involved.  It's internal too so no prying eyes :)

alannajenkins: Like the suggestion of using Confluence Louis, as it means those who are not as famiiar with Github still have access to the information. I've created a page here: https://holidayextras.jira.com/wiki/display/WEB/Incident+Wash-Ups within the Web space. Will update the doc to use that instead.

SkylerAvery:
Once this guideline gets merged in, can we migrate all previously ran incident washups into the confluence page :)

hx-markterry:
Ok with me :horse_racing:

hpoom:
Looks good, but I feel the dropping of "How could it have been avoided" from the [template used up till now](https://github.com/Shortbreaks/about/blob/master/meltdown.md) will mean we don't call for direct action to mitigate incidents in future.

Also we should drop the wording "Wash up" as projects have washup. A blameless postmortem is not a washup, I think postmortems are very different so maybe remove "Wash up" wording from the intro.

Love that we are putting a doc into culture repo, I just think it needs a few minor changes. 

Can we also put .md extension on the end and change the link to [standard markdown link notation](https://guides.github.com/features/mastering-markdown/)

hx-markterry:
@kevinhodges @notABluesSinger @philmeehan Could you review this doc please, as the plan here is to replace the current blameless post mortem document in the shortbreaks/about repo.

@alannamooney could you put a PR into the shortbreaks/about repo to delete the current blameless post mortem document as the group should be using just a single template.

Incident Wash-Up
```diff
@@ -0,0 +1,19 @@
+Incident Wash-Ups should be used to discuss issues that have affected our customers and how we can help to avoid similar situations from occurring in the future. The Wash-Up should include all relevant parties at their earliest opportunity after the event. 
+
+Please use this format for all Incident Wash-Ups so that we have a consistent set of data on how we handle situations. Please add all notes from Incident Wash-Ups to the Confluence space (https://holidayextras.jira.com/wiki/display/WEB/Incident+Wash-Ups), for reference.
```
hx-markterry: It has previously been decided that post mortem notes would be stored in the affected project's repo (see [tripapp post mortems](https://github.com/holidayextras/tripapplite/tree/master/postmortems)), what's the reason for the change here?

hx-markterry:
Thanks for spending the time on this alanna!

hpoom:
I like [point 8's](https://github.com/holidayextras/culture/blob/alannamooney-incident-washup/Incident%20Wash-Up#l19) rewording or point 10 in the original doc, that is an improvement.

Incident Wash-Up
michaelcarter: +1 for storing the post-mortems alongside the platform they've occurred on rather than in confluence.

michaelcarter:
2 points from me: 
- The template looks fine, but I think short breaks and Hx should be working from the same one. Currently SB have their [meltdown](https://github.com/Shortbreaks/about/blob/master/meltdown.md) doc and we have this... What are we doing about ensuring we share the same thing?
- I don't think the writeup should be stored in Confluence, or at the very least shouldn't _only_ be stored in confluence. We should put the write-ups in the repositories of the platforms that they are most associated with. For example, [Tripapp's postmortems folder](https://github.com/holidayextras/tripapplite/tree/master/postmortems).

Incident Wash-Up
alannajenkins: I think storing them in github takes away some of their accessibility - github is not a tool that everyone in our team is familiar with and we shouldn't be expecting everyone to be familiar with it (e.g. Product Owners, Specialists), but these notes should be accessible to everyone. That's why I thought Louis' suggestion of Confluence was a more team-friendly one.

hpoom: The other option would be to put them into Jira. Chat to @notABluesSinger as that is what Shortbreaks are doing right now. Not sure it is ideal, but Confluence does not feel like the right place.

ghost: I think putting it somewhere more accessible in general is a good idea - SB do not use Confluence so perhaps an alternative. Ie creating a JIRA type ticket with a template and link it to the applicable work/feature ticket? Something rolled out across both businesses would be great so we're all consistent.

philmeehan: @notABluesSinger is on holiday for 2 weeks.

I don't think we use confluence - as we have a separate Jira system would this result in an extra cost?

Incident Recovery Jiras are setup as a separate project on our Jira, which has the questions as part of the Jira creation form.

incident recovery

If you have access to our Jira you can see all issues listed here
https://hxshortbreaks.atlassian.net/projects/IR/issues/IR-18?filter=allissues

philmeehan:
I've asked the SB Web Team to have a look at this as well.

I'll have a deeper look when I have cleared down some PRs

hpoom:
@alannamooney Postmortem is [industry standard terminology](https://www.google.co.uk/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=blameless%20post%20mortem). Where we have industry standards I don't think we should look to change them. If we bring new people in or communicate with 3rd parties, or use consultants & contractors then standardised use of terminology will help us out.

Retrospective already has a defined meaning in Agile, and Wash Up is used on larger projects. I think we should stick with Blameless Postmortem. (thats my 2p anyway)

Incident Wash-Up
```diff
@@ -0,0 +1,19 @@
...
+
+    1. Summary
+    
+    2. Affected time period
+    
+    3. Parties involved
+    
+    4. Factors that have contributed
```
lukefrake: Factors that have contributed? What does that mean? Is that the same as say 'What happened?'

If these are possibly being done on a weekend, maybe lets keep them as simple as possible?

Incident Wash-Up
```diff
@@ -0,0 +1,19 @@
+Incident Wash-Ups should be used to discuss issues that have affected our customers and how we can help to avoid similar situations from occurring in the future. The Wash-Up should include all relevant parties at their earliest opportunity after the event. 
+
+Please use this format for all Incident Wash-Ups so that we have a consistent set of data on how we handle situations. Please add all notes from Incident Wash-Ups to the Confluence space (https://holidayextras.jira.com/wiki/display/WEB/Incident+Wash-Ups), for reference.
+
+    1. Summary
+    
+    2. Affected time period
```
lukefrake: Maybe have this as 

Affected time, and sites? We don't have anything down here that tracks blast radius, so we can't tell how much each recovery is talking about.

Incident Wash-Up
```diff
@@ -0,0 +1,19 @@
...
+    
+    3. Parties involved
```
lukefrake: 'Parties involved' could just say 'Who was involved'. 

If I see parties I could put peoples names, or teams, either would make sense IMO. (Never liked the term parties from the SB IR either)

Incident Wash-Up
```diff
@@ -0,0 +1,19 @@
...
+    
+    4. Factors that have contributed 
+    
+    5. Details / known impact
```
lukefrake: Whats the difference between this and point 4?

notABluesSinger:
Hey @alannamooney, what's the current progress with this?
I'd love to take a look once the other points have been addressed.

lizpope:
@notABluesSinger - Alanna is now on Annual leave for a week! I'm going to pick this up in her absence. 

Looking through the comments there seem to be a couple of things outstanding.
1) Where notes from the meeting are kept. Suggestions are confluence, jira and within the repo.
2) The specific points which Luke has raised, which I'll address inline

Incident Wash-Up
lizpope: I think this is what you were asking for in an above point. 2) is time period, 5) is the actual impact (previously blast radius). Perhaps moving one to follow the other would make more sense.

Incident Wash-Up
lizpope: good point, I'd be happy with either 'Parties' or 'Who' - should give same outcome

Incident Wash-Up
lizpope: My suggestion further down would solve this - moving the impact to directly follow the time period.

Incident Wash-Up
lizpope: If I were running this meeting I would take it to mean "what caused the problem" so could be:
no tests so something went live broken or AWS went down or other third party failure or a combination of all those things. Perhaps wording such as "what was the cause" would suit better?

Incident Wash-Up
lizpope: Just having a proper look at this and I like the idea of it being in Jira. But I'm unlikely to be the one running or partaking in the meeting and documentation of it, so would be nice to hear others views...

lizpope:
Are we going to resolve this? If not shall I just close and we continue with the current version? I liked the changes on the new one, but no point if this hanging around.... it's been open for 2 months.

Blameless-Postmortems.md
```diff
@@ -0,0 +1,19 @@
+Blameless postmortems should be used to discuss issues that have affected our customers and how we can help to avoid similar situations from occurring in the future. The Postmortem should include all involved at their earliest opportunity after the event. 
+
+Please use this format for all Blameless Postmortems so that we have a consistent set of data on how we handle situations. Please add all notes from Blameless Postmortem to the Jira project (https://holidayextras.jira.com/secure/RapidBoard.jspa?projectKey=IR&rapidView=332&view=planning.nodetail), for reference.
```
hx-markterry: Can we remove the link as this will only work for our staff and this is a public repo. Maybe just state:

```
Please add all notes from Blameless Postmortem to the `IR Board` Jira project 
```

Blameless-Postmortems.md
```diff
@@ -0,0 +1,19 @@
...
+
+    1. Summary
+    
+    2. Who was involved?
```
hx-markterry: Is this who attended the post mortem or who was involved in the incident? Might be worth clarifying.

hx-markterry:
looking good, only some minor comments from me.

Blameless-Postmortems.md
philmeehan: Would agree with this as it makes sense that SB post-mortems still get logged on our Jira system.

Blameless-Postmortems.md
philmeehan: Both bits of information could be useful :)

philmeehan:
:+1: from me

hpoom:
:+1: from me. Thanks for making the amendments @alannamooney

hx-markterry:
Ok from me

philmeehan:
:shipit: :)

nickloestartup:
:+1: from me - looks good Alanna, makes much more sense. Once merged, do we need to update 'Incident Recovery' on Jira or does that happen automatically?

notABluesSinger:
:shipit:

PR #32, opened 2015-08-24
More links for the technical resources page

hx-markterry:
Please add any links of interest to this doc or in the comments of this PR whichever is easier

PR #34, opened 2015-08-25
Added in more links under existing titles.

hpoom:
Added in more links under existing titles. Also added Sass in as the HX group used both Less and Sass.

I notice there is [another PR](https://github.com/holidayextras/culture/pull/32) open on this file at the mo, so would good to get both :+1:  and merged.


hpoom:
I wonder if we should just link this

https://github.com/ericelliott/essential-javascript-links

hx-markterry:
ok with me :last_quarter_moon_with_face:

PR #38, opened 2015-09-03
Add Components Docs

michaelcarter:
A work in progress, the aim here is to clarify our thinking around what role components play in our user interfaces for the entire team. This **does not** deal with integration issues on specific projects, those need to be documented per-project.

### DIFF
--- a/working-with-components.md
+++ b/working-with-components.md
@@ -38,11 +38,13 @@ Going back to the atomic design principles from the "What are components?" secti
  - **Atom and Molecule Components:** These go in the public [UI-Toolkit](https://github.com/holidayextras/ui-toolkit) project where they can be reused everywhere. This is because at the granular atomic and molecular level, components probably have applicability across all of the Hx platforms as well as to the wider world. In support of this, the UI-Toolkit project is open source.
  - **Organism Components:** These will sit in a separate private [UI-Components](https://github.com/holidayextras/UI-Components) project. This is because these components will have applicability mainly within Holiday Extras' products, and will be subject to rapid change and creation of lots of variants of a similar component for split testing purposes as well as the inclusion of Hx-only code, [react props](https://facebook.github.io/react/docs/reusable-components.html) and the like.
 
-If you're building a new component and you're unsure about where it should go, talk to one of the SAs or the UI-Toolkit team.
+If you're building a new component and you're unsure about where it should go, talk to one of the SAs or the UXUI team.
 
 ## Where can I read more?
 
  - **React**
   - [React Docs](http://facebook.github.io/react/)
+  - [UI Toolkit](http://hungrygeek.holidayextras.co.uk/ui-toolkit/)
+  - [Additional documentation on technical resources](https://github.com/holidayextras/culture/blob/master/technical-resources.md)
  - **React & Tripapp**
   - [Integrating React Components Docs (yet to be written)](#)
--- a/working-with-components.md
+++ b/working-with-components.md
@@ -1,13 +1,13 @@
 # Working with components
 
 ## What are components
-Components can be thought of as sections of user interface on our websites. There are small components, such as buttons and inputs, and larger components such as product tiles and search headers. Typically larger components will be _composed_ of many smaller components. This pattern of designing many small components to compose together for larger reuseable components within a page is known as **atomic design**. To read more on Atomic Design see [Brad Frost's excellent post](http://bradfrost.com/blog/post/atomic-web-design/), but for the purposes of working with components at Hx we'll be referring to three of the 5 atomic design component levels:
+Components can be thought of as sections of user interface on our websites. There are small components, such as buttons and inputs, and larger components such as product tiles and search headers. Typically larger components will be _composed_ of many smaller components, but anything can be a component really. Designing many small components to compose together for larger reuseable components within a page is known as **atomic design**. To read more on Atomic Design see [Brad Frost's excellent post](http://bradfrost.com/blog/post/atomic-web-design/), but for the purposes of working with components at Hx we'll be referring to three of the 5 atomic design component levels:
 
  - **Atoms:** These are the basic building blocks of our pages. Atoms are our most simple components, such as a form label, a text input or a button.
  - **Molecules:** These are groups of atoms bonded together and form larger reuseable components. These molecules take on their own emergent properties and can in turn be reused to build even more complex sections of a user interface. An example of a molecule component could be a label and input together.
  - **Organisms:** These are groups of molecules joined together to form a relatively complex, distinct sections of our user interface. An example of an organism component may be a product tile, or a search header.
 
-At Holiday Extras we use Facebook's [React](https://facebook.github.io/react/index.html) for building composeable user interface components.
+At Holiday Extras we currently use Facebook's [React](https://facebook.github.io/react/index.html) for building new composeable user interface components, but these same atomic design principles apply equally well to other templating languages we use elsewhere - haml, handlebars, etc. 
 
 ## Why should we use components?
 Building a library of reuseable components has several benefits:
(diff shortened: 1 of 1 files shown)


working-with-components.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Where do we put shared components?
+Going back to the atomic design principles from the "What are components?" section we'd like to put our shared components in the following places:
+
+ - **Atom and Molecule Components:** These go in the public [UI-Toolkit](https://github.com/holidayextras/ui-toolkit) project where they can be reused everywhere. This is because at the granular atomic and molecular level, components probably have applicability across all of the Hx platforms as well as to the wider world. In support of this, the UI-Toolkit project is open source.
+ - **Organism Components:** These will sit in a separate private [UI-Components](https://github.com/holidayextras/UI-Components) project. This is because these components will have applicability mainly within Holiday Extras' products, and will be subject to rapid change and creation of lots of variants of a similar component for split testing purposes as well as the inclusion of Hx-only code, [react props](https://facebook.github.io/react/docs/reusable-components.html) and the like.
+
+If you're building a new component and you're unsure about where it should go, talk to one of the SAs or the UI-Toolkit team.
```
jodiedoubleday: I'd say UXUI team here :)

working-with-components.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Where can I read more?
+
+ - **React**
+  - [React Docs](http://facebook.github.io/react/)
+ - **React & Tripapp**
+  - [Integrating React Components Docs (yet to be written)](#)
```
jodiedoubleday, line 50: Maybe a link to the UI-Toolkit (www.ui-toolkit.com)

working-with-components.md
```diff
@@ -0,0 +1,48 @@
...
+Going back to the atomic design principles from the "What are components?" section we'd like to put our shared components in the following places:
+
+ - **Atom and Molecule Components:** These go in the public [UI-Toolkit](https://github.com/holidayextras/ui-toolkit) project where they can be reused everywhere. This is because at the granular atomic and molecular level, components probably have applicability across all of the Hx platforms as well as to the wider world. In support of this, the UI-Toolkit project is open source.
+ - **Organism Components:** These will sit in a separate private [UI-Components](https://github.com/holidayextras/UI-Components) project. This is because these components will have applicability mainly within Holiday Extras' products, and will be subject to rapid change and creation of lots of variants of a similar component for split testing purposes as well as the inclusion of Hx-only code, [react props](https://facebook.github.io/react/docs/reusable-components.html) and the like.
+
+If you're building a new component and you're unsure about where it should go, talk to one of the SAs or the UI-Toolkit team.
+
+## Where can I read more?
```
hx-markterry, line 43: Link to the the technical resources doc under the react heading? (we may need more stuff there)

working-with-components.md
```diff
@@ -0,0 +1,48 @@
+# Working with components
+
+## What are components
+Components can be thought of as sections of user interface on our websites. There are small components, such as buttons and inputs, and larger components such as product tiles and search headers. Typically larger components will be _composed_ of many smaller components. This pattern of designing many small components to compose together for larger reuseable components within a page is known as **atomic design**. To read more on Atomic Design see [Brad Frost's excellent post](http://bradfrost.com/blog/post/atomic-web-design/), but for the purposes of working with components at Hx we'll be referring to three of the 5 atomic design component levels:
```
hpoom: I think we should also use an example of none text based component. Like Map etc.

Maybe not needed, but I would like to get across the message, everything can be a component.

working-with-components.md
```diff
@@ -0,0 +1,48 @@
...
+
+ - **Atoms:** These are the basic building blocks of our pages. Atoms are our most simple components, such as a form label, a text input or a button.
+ - **Molecules:** These are groups of atoms bonded together and form larger reuseable components. These molecules take on their own emergent properties and can in turn be reused to build even more complex sections of a user interface. An example of a molecule component could be a label and input together.
+ - **Organisms:** These are groups of molecules joined together to form a relatively complex, distinct sections of our user interface. An example of an organism component may be a product tile, or a search header.
+
+At Holiday Extras we use Facebook's [React](https://facebook.github.io/react/index.html) for building composeable user interface components.
```
hpoom: I think React is the current goal, but we can achieve components and atom design with jump simple template engines like handlebars etc. We should encourage atomic design even on old templating platforms. Also in a few years we would probably have moved on from React (something new will come along)

working-with-components.md
```diff
@@ -0,0 +1,48 @@
...
+This should be componentised from the beginning, and this is the expectation for new pull requests. The UX/UI Team are on hand to offer advice and assistance with this if necessary.
+
+### Modifying existing user interface code
+When modifying existing user interface code, existing user interface elements that can be replaced by shared components we already have should be replaced. If there are no existing user interface components we should look at building some when time allows.
+
+
+## Where do we put shared components?
+Going back to the atomic design principles from the "What are components?" section we'd like to put our shared components in the following places:
+
+ - **Atom and Molecule Components:** These go in the public [UI-Toolkit](https://github.com/holidayextras/ui-toolkit) project where they can be reused everywhere. This is because at the granular atomic and molecular level, components probably have applicability across all of the Hx platforms as well as to the wider world. In support of this, the UI-Toolkit project is open source.
```
hpoom: " components probably have applicability across all of the Hx platforms as well as to the wider world. In support of this, the UI-Toolkit project is open source."

I don't like the use of the word _probably_, sorry but it does allows for ambiguity. Also we should probably not state HX, and just tates that:
`This is because at the granular atomic and molecular level, components have applicability across all uses in the web, they are not specific to HX and could be used by any company much the same way Bootstrap is not Twitter specific. In support of this, the UI-Toolkit project is open source.`

hpoom: My wording is crap, but you get the point. Leverage an example like bootstrap. Make is clear that if your about to add a component and it is very HX specific it does not go in Atoms or Molecules

working-with-components.md
hpoom: I would agree

working-with-components.md
hpoom, line 50: Link to all relevant repos.

Also in time this doc need to link to an example of how to embed components in another project/system

hpoom:
Overall a great doc. We should get some other FED eyes on this too from a group perspective. @lukefrake @MCooke can you have a quick read. Could/Should this apply to Shortbreaks?

I am thinking group tech here, any feedback on this doc to make it applicable across the group? For me it looks OK as it is and I can't see why we can't follow these guidelines.

MCooke:
I love the idea of explaining components and our procedure around using them. Atomic Design is a solid way of representing components but we tend to steer clear of it. We generally describe components using BEM methodology. http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

Incorporating components in front-end PR's is fantastic, would be great to see examples.

If the HX specific language was removed or changed it would be better for our use, such as UX/UI and SA.

The last section is too specific to React as an implementation of components. Components exist without React and i'd suggest separating the idea of components from any implementation.

hpoom:
Thanks @MCooke I agree we should change `The last section is too specific to React as an implementation of components. Components exist without React and i'd suggest separating the idea of components from any implementation.`

That way when the technology changes this doc still holds true.

On the BEM stuff, can we not have that and Components? Are Components not markup (HTML) and the events related to that markup (JS), and BEM used for styling and how we apply brand to components and templates? 

_Maybe I have it wrong, but would love to understand it. Also conscious we get this PR in a good state, get it merged and then apply other feedback on future PRs on this doc._

michaelcarter:
Thanks for the feedback all. I've gotta put this aside for the rest of today but will pick it back up and address the rest of these comments on Monday.

MCooke:
BEM is a naming convention for CSS and HTML but it's also used to describe components and their constitute parts. More than happy to sit down and show you how we use it.

I think its best to push this PR through for you guys. Then we can talk and see if we can make this group wide.

michaelcarter:
@MCooke Would you mind if we left the BEM vs. atomic design as a conversation for another day? The team our end are invested in atomic design, and my main priority here is actually getting people using components.

michaelcarter:
I'm going to merge this PR tomorrow morning unless anyone else has stuff they'd like to add! After that feel free to make any suggestions via a PR of your own.

MCooke:
@michaelcarter 100%, great just to get people thinking about components. Was just my two cents.

PR #41, opened 2015-09-18
Change assert to sinon.assert

rahulpatel:
Using `sinon.assert` to test sinon functions provides much better error messages if that assert ever fails.

```
> assert.ok(someModule._add.calledWith('first', 'third'));
AssertionError: false == true
```

```
> sinon.assert.calledWith(someModule._add, 'first', 'third');
AssertError: expected add to be called with arguments first, third
```

### DIFF
--- a/testing-standard.md
+++ b/testing-standard.md
@@ -39,7 +39,7 @@ sinon.stub(someModule, '_add').returns(expectedResult);             // stubbing 
 var result = someModule._proxy(first, second);
 
 assert.equal(result, expectedResult);                               // assert the return value
-assert.ok(someModule._add.calledWith(first, second));               // assert the stub was invoked as expected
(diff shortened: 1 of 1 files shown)


connormeredith:
I agree, the output from sinon is much more informative :+1: 

Sinon documentation on assertions: http://sinonjs.org/docs/#assertions-api

hx-markterry:
I don't think its that sinon is more informative, its because we are abusing the use of `assert.ok` I think this should only be used for comparing a boolean value. The base assertion library only offers basic comparisons and yes they can be used for everything but its not nice. Using a additional library with more specific assertions is the way forward. In other projects we use `chai` to provide many additional assertions which fit the tests required.

pmcnr-hx:
:+1: from me.
@hx-markterry: this is restricted to `sinon` related assertions, and does seem to be orthogonal/complementary to using a richer assertions library instead of the Node standard library one.

PR #42, opened 2015-09-21
added recommendations

hx-markterry:
Added which providers we should be using and reiterated not passing secret keys to them 
- [x] It's ok
- [x] I agree with the other guy


shackpank:
:+1:

rahulpatel:
:+1: Seems sensible

PR #45, opened 2015-10-08
Update unit testing standards

michaelcarter:
WIP

### DIFF
--- a/testing-standard.md
+++ b/testing-standard.md
@@ -10,8 +10,30 @@ In order to keep test writing familiar across the Holiday Extras group, we shoul
 * [Chai](http://chaijs.com/): For making assertions in our tests
   * [Sinon-chai](https://github.com/domenic/sinon-chai): For integrating Sinon with Chai
   * [chai-as-promised](https://github.com/domenic/chai-as-promised/): For clean promise assertions
-  * With Chai, we always use the `expect` syntax for our tests.
 * [React Test Utilities](https://facebook.github.io/react/docs/test-utils.html): Helpers for testing React components
+
+### Assertion Syntax
+With Chai, we always use the `expect` syntax for our tests. By keeping a single assertion syntax across the Holiday Extras group we're making it easy for developers to get going with writing tests on any of our projects. Chai + expect provides a clean, readable style of test writing, for example:
+
+```javascript
+expect(foo).to.be.a('string');
+expect(foo).to.equal('bar');
+expect(foo).to.have.length(3);
+expect(beverages).to.have.property('tea').with.length(3);
+```
+
+And with the addition of sinon-chai for sinon stubs:
+
+```javascript
+expect(myStub).to.have.been.calledWith('foo');
+```
+
+
+### Linting Issues
+You may occasionally encounter a slight issue here with our linting, specifically our linting complaining about unused expressions in the tests due to Chai's syntax. There are a couple of viable workarounds for this:
+
+ * Turn off `no-unused-expression` linting rule for your tests.
+ * Use a library like [chai-lint](https://www.npmjs.com/package/chai-lint) to provide functions that replace the property getters provided by Chai.
 
 ## 1.2 Good Test Folders
 When addind new tests to projects, test files should sit in a folder hierarchy that mimics the hierarchy fo the source code. The test files themselves should be named similarly to the source files, but with `test` at the beginning of the file name. For example, if we have a source file `src/views/checkoutView.js` we should have a test file `test/unit/views/testCheckoutView.js` or similar. This makes finding tests an easy and logical process.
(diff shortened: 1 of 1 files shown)


testing-standard.md
```diff
@@ -1,57 +1,112 @@
...
+  
+  	  var testFirstName;
+  	  var testLastName;
+  
+  	  beforeEach(function() {
+  	    testFirstName = 'Geoffrey';
+  	    testLastName = 'Banks';
+  	  });
+
+	  context('when no first name is provided', function() {
```
hxoliverrumbelow: We've got some mixed tabs + spaces here throwing it out a little

hxoliverrumbelow:
Looks good :+1:

testing-standard.md
```diff
@@ -64,23 +119,70 @@ someModule._action = function(number) { // test with `null`
...
+* Reference variables for expected results to ease duplication.
+
+### Within `it` blocks:
+
+* Check all return/callback values
+* Check every stub is invoked with all expected parameters
+
+## 1.6 Unit Testing Gotchas
+
+### Hard to test functions
```
kevinhodges, line 210: this should probably be in our server side best practices, about single responsibility principle and keeping functions small and this doc should link to that for it's example with a brief explanation here.

testing-standard.md
```diff
@@ -1,57 +1,112 @@
...
+  * [Sinon-chai](https://github.com/domenic/sinon-chai): For integrating Sinon with Chai
+  * [chai-as-promised](https://github.com/domenic/chai-as-promised/): For clean promise assertions
+* [React Test Utilities](https://facebook.github.io/react/docs/test-utils.html): Helpers for testing React components 
 
-### Each test should:
-* Use product codes that are guaranteed to be available
-* Use dynamic dates that are a fixed amount of time in the future
-* Not assert against any dynamic data (eg. content) that may change unexpectedly
+## 1.2 Good Test Folders
+When addind new tests to projects, test files should sit in a folder hierarchy that mimics the hierarchy fo the source code. The test files themselves should be named similarly to the source files, but with `test` at the beginning of the file name. For example, if we have a source file `src/views/checkoutView.js` we should have a test file `test/unit/views/testCheckoutView.js` or similar. This makes finding tests an easy and logical process.
```
joestump: Small quibble, but most test runners fine tests with a regular expression; it's conceivable that actual code could start with `testFoo` (e.g. a framework for A/B testing or a model called `Test`). I'd vote for a hyphen between `test` and the actual file name.

testing-standard.md
```diff
@@ -90,14 +192,17 @@ someModule._action = function(number) {
...
 ```
-becomes:
-```
+
+Becomes:
+
+```javascript
 someModule._numberLoop = function(number) {
   if (i%3 == 0) number++;
   if (i>4) number++;
```
joestump: Do our linting standards allow for this kind of spacing?

testing-standard.md
joestump, line 210: This section should make mention of the [dependency injection pattern](https://en.wikipedia.org/wiki/Dependency_injection).

testing-standard.md
hx-markterry: I'd vote for not adding the word `test` to a filename, we and test runners know it is a test due to being in the `test` directory.

testing-standard.md
```diff
@@ -105,3 +210,38 @@ someModule._action = function(number) {
...
+# 2.0 Selenuim Tests
+
+TODO
+
+# 3.0 System Tests
+
+## Each test should:
+* Use product codes that are guaranteed to be available
+* Use dynamic dates that are a fixed amount of time in the future
+* Not assert against any dynamic data (eg. content) that may change unexpectedly
```
joestump: You're dancing around the topic a bit here, but we should specifically call out the need to fixture data and that the unit test suite should be capable of being ran without any backing services.

testing-standard.md
hx-markterry: no, also other examples in this doc, do not pass linting.

testing-standard.md
hxoliverrumbelow, line 210: To be honest, we tried the dependency injection pattern in our serverside projects, then we backtracked and took most of it out. We found that most things can exist as singletons, and as such can be `require()`d in nicely. I can think of at least one of our big projects where we're dependency injecting things that we shouldn't be.

joestump, line 210: Oh, oh God. Singletons? Shared global state is a sad, sad path. http://blogs.msdn.com/b/scottdensmore/archive/2004/05/25/140827.aspx

testing-standard.md
michaelcarter: No, sorry. These are old. I'll update.

testing-standard.md
pmcnr-hx, line 210: My 2 pennies: with Javascript, especially server side with NodeJS, I end up solving most problems with modules, functions and plain objects. When everything doesn't need to be an object, the _Singleton_ vs _Dependency Injection_ dilemma might not even pose itself. Modules are a wonderful abstraction, implementing a closure and namespacing the symbols defined within, so a module level variable is in my opinion a perfectly acceptable solution for state that needs to be shared (database connection pools, loggers, caches, ...).

hxoliverrumbelow, line 210: That article was written a long time ago for people who code in OO languages, it's written for languages that require objects to be constructed and languages that are difficult to control/test without some kind of DI. Very few of our NodeJS modules require construction and the NodeJS module pattern makes it easy to stub/control/test code and identify dependent modules.

If you strongly disagree, lets have a video chat about it and involve some of the other SAs. I can provide plenty of examples where we used to have it and then removed it to reduce complexity. I can also bring up the core part of one of our frontend projects using DI from before we had requireJS or Browserify - now we've solved those challenges the DI is actually making things worse compared to `require()`ing in the "singletons" (eg, we only ever have one 'session' model).

testing-standard.md
michaelcarter: I think I'm pro removing `test` altogether, given what @hx-markterry has pointed out. Any objections?

PR #50, opened 2015-10-16
Let's start using ES6

Joezo:
#### What does this PR do?

Enables ES6 features in our `.eslintrc` file and updates the docs surrounding ECMAScript usage. 
## 
- [x] I like this
- [x] I like this as well


hxpaul:
I like this, but think this might need some "i am an architect and i like this" buttons.
I like this because we're already relying on jsx already, which needs transpiling.
This is not another coffeescript...

hx-markterry:
I've added it to the SA meeting agenda for discussion later today.

pmcnr-hx:
I welcome our ES6 (ES2015?) overlords: :+1:

lukefrake:
Yay ESFuture!!! Nice work @Joezo :+1: 

Is it worth mentioning transpilers in here, I assume you will start using Babel?

Joezo:
yup, in the changes it mentions Babel and Node 4 :)

lukefrake:
@Joezo sorry, I meant in `technical-resources.md` :jack_o_lantern:

rahulpatel:
I'm so for this! :+1:

.eslintrc
```diff
@@ -13,7 +13,29 @@
...
-    "jsx": true
+    "jsx": true,
+    "arrowFunctions": true,
+    "binaryLiterals": true,
+    "blockBindings": true,
+    "classes": true,
+    "defaultParams": true,
+    "destructuring": true,
+    "forOf": true,
+    "generators": true,
```
hx-markterry: @hxoliverrumbelow did you mention in the SA meeting that you would not want to allow all es6 features, I think `generators` was one you weren't keen on?

hx-markterry:
yey we now get block scoped vars!

.eslintrc
pmcnr-hx: I'm interested in the rationale for disallowing any ES2015 features. In my experience generators are a very useful abstraction.

Joezo: I agree with pedro, i'm of the opinion we let everything through for now. If we later decide a feature isn't good for us then we pull it.

hxoliverrumbelow: @hx-markterry There are indeed a whole bunch I don't want us using - that's not to say we should lint against them, there might be some exceptional use cases, but as far as daily coding goes, we need to be standing by our best practices. @Joezo is putting forward the bits he feels strongly about, then if we're all in agreement we'll whitelist those features and train/support our dev team in implementing them. If someone uses a feature that isn't mentioned in the best practices doc, expect it to get slammed in a PR (eg generators, let, es6 module syntax) just like any other bad practice.

hxoliverrumbelow: @pmcnr-hx @Joezo just to clarify - allowing the linter to pass things like generators does NOT mean we'll be allowing them to be used in production.

hxoliverrumbelow:
@hx-markterry not quite ;)

.eslintrc
hxpaul: if it's not going to be allowed then blocking it in the linting for now seems sensible, it would avoid any confusion

hx-markterry:
Before this PR moves forward, it needs to contain an update to our javascript best practices to mention which features we care / dont care about. Maybe adding joes gist to the doc: https://gist.github.com/Joezo/f979aeb7ffb4d9d69fac would be a good start.

This will make the proposed changes more visible and then everyone can discuss the proposed feature set which seems to be the main issue here.

Once we have that agreed, then we can do one of two things (also to be discussed):

1) Only allow recommended features to pass linting:

or 

2) Just add the `es6` environment to the lint rule and remove the specific rules, which i think will allow everything and be clearer

.eslintrc
pmcnr-hx: Agree with @hxpaul: if it violates best practices and can be caught by the linter, then it should be enforced by the linter. We should automate anything that can be automated.

Joezo:
I'm of the opinion that we enable everything. I don't think guidelines need updating, people should use they're knowledge to determine whether they _need_ to use generators etc. There is nothing in our guidelines mentioning specific language features we shouldn’t use in ES5. I think it’s all about using common sense. Just to clear things up, this PR's intention was to enable the next version of JS to be used. It wasn't about championing certain features etc. I'm going to update the eslint file to use the `es6` environment rule that @hx-markterry mentioned as I didn't know that existed :)

lukefrake:
We're using babel on our new ATG project really heavily. 

We've been using `classes`... YAY! Something to bare in mind though, babel doesn't support IE8, you have to use this [loose](https://babeljs.io/docs/advanced/loose/#usage) property if you want to use `classes` (and some other stuff I'm sure), which will mean some things will work. But some will not.

As lots of people are reading this thought I would drop it here as I spend a few hours today thinking WTH!

Joezo:
In the interest of getting us using something i've shrunk the feature set down to the ones which _I_ can't see anyone having a problem with. Notable exclusions are classes, generators and modules. The features that are included are: 
https://leanpub.com/understandinges6/read#leanpub-auto-computed-property-names
https://leanpub.com/understandinges6/read#leanpub-auto-property-initializer-shorthand
https://leanpub.com/understandinges6/read#leanpub-auto-arrow-functions
https://leanpub.com/understandinges6/read#leanpub-auto-block-bindings
https://leanpub.com/understandinges6/read#leanpub-auto-destructuring
https://leanpub.com/understandinges6/read#leanpub-auto-rest-parameters
https://leanpub.com/understandinges6/read#leanpub-auto-the-spread-operator
https://leanpub.com/understandinges6/read#leanpub-auto-template-literals

hxoliverrumbelow:
### No brainers

https://leanpub.com/understandinges6/read#leanpub-auto-arrow-functions
https://leanpub.com/understandinges6/read#leanpub-auto-template-literals
https://leanpub.com/understandinges6/read#leanpub-auto-block-bindings
These all look great, full adoption across the board would be a good thing.

### Auto Computed Property Names

https://leanpub.com/understandinges6/read#leanpub-auto-computed-property-names

```
var suffix = " name";
var person = {
    ["first" + suffix]: "Nicholas",
    ["last" + suffix]: "Zakas"
};
console.log(person["first name"]);      // "Nicholas"
```

This one smells really bad to me. There are two ways of storing complex dynamic data and manipulating it:

```
// Option 1:
var data = {
  "davesFavourite": 12,
  "olisFavourite": 7,
  ...
};
Object.keys.map(function(i) { return data[i]; }).map(function(i) { /* yay */ });
for (var key in data) { var i = data[key]; /* yay */ };

// Option 2:
var data = [
  { name: "dave", "favourite": 12 },
  { name: "oli", "favourite": 7 },
];
data.map(function(i) { /* yay */ });
```

From all my experiences in HAPI and with hxMetrics (which is a MINEFIELD of this type of dataset), the second approach trumps the first approach. As such, any kind of shorthand to make the first approach easier to achieve is a bad thing and should be avoided.

I'm against this one. It'd be a green light to adopt bad practices.

### Auto Property Initializer Shorthand

https://leanpub.com/understandinges6/read#leanpub-auto-property-initializer-shorthand

```
function createPerson(name, age) {
    return {
        name,
        age
    };
}
```

My main concern here is that we're only a couple of pixels away from this:

```
function createPerson(name, age) {
    return [
        name,
        age
    ];
}
```

and both would be syntactically valid, but one is horribly broken. I can't think of many occasions where we'd benefit from this, I'm not sure if the pain would be worth it.

I'm on the fence on this one.

### Auto Destructuring

https://leanpub.com/understandinges6/read#leanpub-auto-destructuring

```
var options = {
  repeat: true,
  save: false,
  rules: {
    custom: 10,
  }
};

// later

var { repeat, save, rules: { custom }} = options;

console.log(repeat);        // true
console.log(save);          // false
console.log(custom);        // 10
```

I've got 2x concerns with this one.
1. It'd be easy to stop being explicit about what data is being passed around. Right now our serverside best practices prefer us to name every property we ship between modules. When someone works on code that is explicit, they don't need to jump around pushing in `console.log`s everywhere attempting to guess what object properties they need, or if the set of properties they got are guaranteed or not. This rule would make it easier, possible acceptable, to drop that best practice.
2. The case for destructuring function return values (for example, returning a tuple), can be more of a BAD practice than a good one. Consider this function:

```
function getData() { return [ "oliver", 27 ]; }
```

As a consumer of this function I'll get an array of two values. Without looking at the code, or hunting for documentation, I've got no idea what those data items mean. After some digging I might be able to destructure it like so:

```
var name, age;
[name, age] = getData();
```

and from then on, all future consumers of the original function will have to go through the same process of "what does this data mean?!"

The far better practice, that we currently implement, is whereby we always return a named object:

```
function getData() {
  return {
    name: "oliver"
    age: "27"
  };
}
```

any consumer of this module can very quickly tell what data is coming back, and without looking at the code, just looking at the return value, it's extremely obvious what data they've got.

I'm against this one. It'd be a green light to adopt bad practices.

### Auto Rest parameters + Spread operator

https://leanpub.com/understandinges6/read#leanpub-auto-rest-parameters
https://leanpub.com/understandinges6/read#leanpub-auto-the-spread-operator

```
function pick(object, ...keys) {
    let result = Object.create(null);

    for (let i = 0, len = keys.length; i < len; i++) {
        result[keys[i]] = object[keys[i]];
    }

    return result;
}
```

```
let values = [-25, -50, -75, -100]

console.log(Math.max(...values, 0));        // 0
```

This one is similar to destructuring. We are currently always explicit in naming function arguments and reducing the params to each function:

```
var doStuff = function(name, list) { /* ... */ };

var name = "oli";
var list = [ 1, 4, 5, 7, 9 ];
doStuff(name, list);
```

I don't see this syntactic sugar providing any good reason to change this practice - ultimately, our arrays of data will always be arrays, databases will always callback with `err, dataList`. API calls will always return lists of objects. It feels like this feature was added to enable some more extreme functional styles of programming.

I'm against this one.

michaelcarter:
I think there's definitely potential to misuse these language features as @hxoliverrumbelow has highlighted, but that at the end of the day programming has a huge shoot yourself in the foot potential by its very nature.

Sooner or later these will be supported natively across ES programming platforms (browsers, serverside etc.), and we'll just have to deal with their existence whether we individually agree they're good for us or not. 

I think this pretty much forces the adoption issue to one of education - i.e. "What anti-patterns will could I run into with my newfound ES6 power?". Oli's done a great job of highlighting some of these, so I think we should just get those into our JS best practices docs and then enable these features "embracing the future" of our stack's language, rather than putting this off. In the interim (while we document what good and bad usage looks like for some more complex ES6 features) merge the no-brainers and get those out of the way.

What say you, @hxoliverrumbelow, @Joezo?

Joezo:
I like the approach @michaelcarter. To clarify, we should add Oli's anti-patterns to our best practices doc and then enable the features. This gives clear guidance on how not to use those features and leaves space for people to discover a good use of them in future. @hxoliverrumbelow are you happy for me to add your examples to our docs as part of this pull request?

hxoliverrumbelow:
@Joezo @michaelcarter Sure, lets give it a go :+1:

hxoliverrumbelow:
:+1:

hx-markterry:
ok with me :key:

PR #52, opened 2015-10-21
Functional programming 👉 partial functions

pmcnr-hx:
Some of us are partial (pun intended) to the functional programming paradigm. This can be a bit confusing for those who have less exposure to the _functional_ way of doing things. We don't claim that _functional_ is better than _imperative_ (although it mostly is) but try instead to clarify the functional alternatives to the more traditional constructs.
- [x] :+1: 
- [x] :+1: :+1:

### DIFF
--- a/general-javascript-best-practices.md
+++ b/general-javascript-best-practices.md
@@ -79,6 +79,66 @@ We avoid this issue in the good examples. Depending on the complexity of the con
 
 **Caveat:** This approach isn't always practical with more complex functions. For example, a function may contain complex conditional logic that can't be succinctly reflected in its name, and that we don't want to inline and repeat everywhere. In this scenario, we can use things like throwing `Errors` or `true`/`false` return values for sync code and rejecting promises & erroring callbacks for async code. The ultimate aim is just to ensure that the programmer isn't mislead about the behaviour of the function, and encourage them to consider handling non-happy path behaviour.
 
+
+## Functional JavaScript
+
+JavaScript's
+[first-class functions](https://en.wikipedia.org/wiki/First-class_function)
+enable it to support the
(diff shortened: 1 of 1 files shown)


serverside-javascript-best-practices.md
```diff
@@ -680,6 +680,65 @@ async.map(['http://foo', 'http://bar'], dbInfo, ...);
...
+```
+
+The following are all equivalent ways to implement the `subtractFrom5` and
+`subtract5` functions using `subtract`:
+
+* Wrapper functions:
+
+  ```javascript
+var subtractFrom5 = function (x) {
+  return subtract(5, x);
```
hx-markterry: There is a problem with the indentation here, when the markdown is viewed on github.

hx-markterry:
I agree with the above, but it should be moved to the general javascript best practices document as it is not server specific.

pmcnr-hx:
@hx-markterry: I've added here because @hxoliverrumbelow mentioned we would be merging the general javascript best practices document with this one. Will we actually be doing that sometime soon or will these always be separate documents and we'll just move part of the server-side best practices that apply generally to the general document?

hx-markterry:
I'm sure there will be some server or client specific parts of the javascript best practices, how it is arranged does not really matter as long as it is clear. My main concern here is that the best practice suggested above applies to both and should be under the general document, rather than deferring the merging of practices to someone else or another point in time.

pmcnr-hx:
Will move to general @hx-markterry.

pmcnr-hx:
@hx-markterry: moved to general best practices.

hx-markterry:
ok with me :monkey:

pmcnr-hx:
This has now been rebased on current master.

philmeehan:
:+1:

PR #54, opened 2015-10-22
Missing element

robhuzzey:

Adds a little more docs on recent findings around React.

N/A
#### How can this be tested?

Read it.
#### Screenshots / Screencast

N/A

N/A
#### Developer Definition of Done/Quality Checklist (for PR author to complete BEFORE code review):
- [x] I have checked the [contributing document](../blob/master/CONTRIBUTING.md) and I'm happy for this to be reviewed.
#### Software Engineer or Developer review:
- [ ] I’ve witnessed the work behaving as expected (this could be on the authors machine or screencast).
- [ ] I’ve checked for coding anti-patterns.
- [ ] I’ve checked for appropriate test coverage.
- [ ] I’ve checked all the tests are passing.
#### Software Engineer or project guru review:
- [x] I’ve witnessed the work behaving as expected (this could be on the authors machine or screencast).
- [x] I’ve checked for coding anti-patterns.
- [x] I’ve checked for appropriate test coverage.
- [x] I’ve checked all the tests are passing.

### DIFF
--- a/git-best-practices.md
+++ b/git-best-practices.md
@@ -13,6 +13,7 @@ the branches from accidentally:
 * Being force pushed
 * Being deleted
 * Having changes merged into them until required status checks pass
+  * __not recommended__ - this requires all pull requests to be up to date with master to be mergeable. Updating is a manual process even if an automatic merge is possible and it adds a lot of additional build workload.
 
 ## Branching
 


react-troubleshooter.md
```diff
@@ -4,6 +4,16 @@
 
 ### __invariant violation__:
 
+#### Unable to find element
+
+When the DOM has mutated & you try to modify props / state on a component, you will find that React will complain about not being able to find the element.
```
lukehansell: is it specifically when the DOM above a component has been modified to remove the component's DOM or any DOM manipulation? DOM DOM DOM DOM...

react-troubleshooter.md
```diff
@@ -4,6 +4,16 @@
...
+
+Usually, we would expect `React.Component.componentWillUnmount` to be triggered before the component is unmounted from the DOM or `React.Component.componentDidUnmount` to trigger after but this does not happen if the DOM element is destroyed by other methods outside or React's control (`document.body.innerHTML` for example).
```
lukehansell: `componentDidUnmount` doesn't exist

hxpaul:
taking a review... I think there's a different template for this one just "I like this" and "I also like this", but yes I agree. This was a great find Rob,. :+1:

react-troubleshooter.md
```diff
@@ -4,6 +4,16 @@
...
+
+React doc's comment about this with AJAX calls [here](https://facebook.github.io/react/tips/initial-ajax.html) is relevant as this issue can affect ajax calls that callback after the DOM has mutated so one fix is to use `React.Component.isMounted()` however if the DOM was completely removed, you may not even have access to `React.Component` anymore.
+
+Where we may encounter potential issues like this, it is better to wrap any functions that modify state or props in a try / catch to ensure the application doesn't fail in these edge cases.
```
lukehansell: we should never be modifying props

react-troubleshooter.md
michaelcarter: DOM DOM DOM

react-troubleshooter.md
robhuzzey: Yeah, badly worded, I meant if we pass in different props. An example of this is when you change a prop & the `componentWillReceiveProps` method is called. Maybe I should state that if we pass in new props we should protect in that method?... or in fact should we be that cautious?

react-troubleshooter.md
robhuzzey: I know!... not sure why I wrote that actually? :/

react-troubleshooter.md
robhuzzey: This issue only seems to happen when the DOM that React has "bound to" (for want of a better way to describe it) has gone. This would also apply to parent nodes as the child would be removed at the same time the parent is removed. Essentially, if you change a thing & React has no idea what you've done, it throws this error.

lukehansell: Could you add this above to make the case clearer please

PR #56, opened 2015-10-23
Prefer global `'use strict';' at the module level.

pmcnr-hx:
From version 1.0 ESLint has clear `function` and `global` (defaulting to `function`) settings for the `strict` rule. For NodeJS projects the `global` setting makes the most sense, and now that we've switched to using Browserify for front-end projects, `global` should do the right thing for those also, while avoiding the verbosity of one `'use strict';` directive per function.


shackpank:
:+1:

hxpaul:
also :+1: - I think we've mostly been doing this so good to legislate it

PR #57, opened 2015-10-26
Recommend against requiring status checks on branch protection

shackpank:
Observations from having it switched on in a frequently changing repository for a few days.

cc @pmcnr-hx @hx-markterry as I'm suggesting backing out something you introduced recently...
- [x] 1
- [x] 2


pmcnr-hx:
:+1: I concur. I had initially the "checks before merge" as optional. Preventing accidental forced pushes to `master` and other non-feature branches is the real value of branch protection. Merges can always be reverted without altering history.

Joezo:
Yusss!

PR #59, opened 2015-11-01
New explicit rule set for ESLint 1.8.0.

pmcnr-hx:
This PR updates our recommend ESLint rule set with explicit rules for the recommendations generated by the latest ESLint release.

This is a pre-requisite for upgrading the ESLint version used by make-up which is now over 5 months old.
- [x] :+1: 
- [x] :shipit:


hx-markterry:
Taken

.eslintrc
```diff
@@ -35,29 +37,57 @@
     "no-array-constructor": 2,
     "no-caller": 2,
     "no-catch-shadow": 0,
+    "no-cond-assign": 2,
+    "no-constant-condition": 2,
+    "no-control-regex": 2,
```
hx-markterry: This will include newlines and tabs etc:

```
var pattern = new RegExp("\n");
```

.eslintrc
```diff
@@ -35,29 +37,57 @@
...
+    "no-empty-character-class": 2,
     "no-empty-label": 2,
     "no-eval": 2,
+    "no-ex-assign": 2,
     "no-extend-native": 2,
     "no-extra-bind": 2,
+    "no-extra-boolean-cast": 2,
     "no-extra-parens": [2, "functions"],
+    "no-extra-semi": 2,
+    "no-extra-strict": 2,
```
hx-markterry: This has now been replaces with the rule `strict`

.eslintrc
```diff
@@ -35,29 +37,57 @@
...
     "no-implied-eval": 2,
+    "no-inner-declarations": [2, "functions"],
+    "no-invalid-regexp": 2,
+    "no-irregular-whitespace": 2,
     "no-iterator": 2,
     "no-label-var": 2,
     "no-labels": 2,
     "no-lone-blocks": 2,
     "no-loop-func": 2,
+    "no-mixed-requires": [2, false],
```
hx-markterry: The boolean config of this rule is deprecated according to the docs, this way is preferred:

```
{"grouping": false}
```

.eslintrc
```diff
@@ -35,29 +37,57 @@
...
     "no-nested-ternary": 2,
     "no-new": 2,
     "no-new-func": 2,
     "no-new-object": 2,
+    "no-new-require": 2,
     "no-new-wrappers": 2,
+    "no-obj-calls": 2,
+    "no-octal": 2,
     "no-octal-escape": 2,
+    "no-path-concat": 2,
```
hx-markterry, line 85: YES!

hx-markterry:
Only a couple of points here

pmcnr-hx:
@hx-markterry: updated configuration according to your feedback.

hx-markterry:
ok with me. :balloon: 

I've just tested make-up with the newer version of eslint and its tests don't pass, this is probably due to an eslint API change, going from eslint's rapid development.

pmcnr-hx:
@hx-markterry: like we talked the other day I think it might just be that the recent ESLint versions parser picks up things that were missed by the older version. When I submit the PR for `make-up` I'll fix those issues.

hxpaul:
i'll take this, trying it out now (and then frantically getting all my pr's in before this is merged, obviously)

pmcnr-hx:
:smiley_cat: thanks @hxpaul!

.eslintrc
```diff
@@ -26,6 +27,7 @@
     "dot-notation": [2, { "allowKeywords": true }],
     "eol-last": 2,
     "eqeqeq": 2,
+    "handle-callback-err": [2, "^(err|error)$"],
```
robhuzzey, line 30: Shouldn't we just settle on one rather than `err` or `error`?

hxpaul:
ok by me :+1: :shipit: 
one sort of related point, the make-up npm module links to a ruleset by commit hash by the looks, could / should it check out this repo and always use the master version of this file? Right now to enforce this linting we need to a) update the make-up repo and b) update any repos that are tied to a specific version of the make-up repo

pmcnr-hx:
I think we want to be conservative in our approach @hxpaul and not break linting the passes in a project, so at the moment, we will require a minor version update of `make-up` if the rule set changes. Projects will have to update the `make-up` dependency manual if they wish to use the latest and greatest.

PR #63, opened 2015-11-05
Correct value for `jsx-quotes` rule

pmcnr-hx:
Apologies, I was a bit distracted and didn't update the value from the old `react/jsx-quotes` rule to the one expected by the new `jsx-quotes` rules. This fixes the issue and (finally) correctly updates the rule set for the forthcoming `make-up` ESLint 1.8.0 update PR.
- [x] :+1: 
- [x] :shipit:


DuncanFenning:
Nice :monkey_face:

hxpaul:
:+1:

PR #67, opened 2015-11-12
Document publishing modules to public npm

shackpank:
Some notes on publishing / updating / managing public modules.

I can add more detail on specific points if people feel strongly that we need to, but in general this feels like you could spend a long time doing that. I wanted to get some basics documented as @pmcnr-hx asked how you'd go about doing this (in the context of this organisation) and I realised none of it was written down anywhere.
- [x] At least one Distributioner
- [x] At least one Shortbreaker
- [ ] :memo: :white_check_mark:

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ This is a collection of documents that outline Holiday Extras' culture and devel
  * [General Meetings](general-meetings.md)
  * [Technical Planning Meetings](technical-planning-meeting.md)
  * [Deployment Guidelines](deployment-guidelines.md)
+ * [Publishing to npm](publishing-to-npm.md)
  * [Definition of Ready and Done](definition-of-ready-and-done.md)
 
 ### Other
--- a/publishing-to-npm.md
+++ b/publishing-to-npm.md
@@ -1,0 +1,23 @@
+# Publishing to npm
+
+For any problems that are not specific to the business, we should look to use publicly available modules and release our own when no suitable existing options are available. Many of our projects are javascript, so these modules will probably live on [npm](http://npmjs.com/).
+
+The guidance below covers the absolute basics, some modules may automate this process and if they do should provide details of how to release them in their own README/documentation.
+
+### Who to publish as
+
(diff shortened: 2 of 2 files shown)


pmcnr-hx:
This is exactly what I would have been looking for before talking to you @shackpank: :+1:. Should it be linked from the main README (or another document)?

philmeehan:
Hi @shackpank,

I haven't published myself so I can't comment on the correctness of the process but the explanations do look good so I'll :+1: 

Few Qs related to npm in general, may be worth running them past in the group tech catch-up:

1) think we have a couple of modules published under Shortbreaks, may be worth folding them in with holidayextras:
https://www.npmjs.com/package/praetorian
https://www.npmjs.com/package/iin-checker

2) I suggested to a couple of people a while ago that we should maybe create a 'Hungry Geek' account for our open source code, which I think a few organisations do (like Asda / Spumko / Hapijs). Then we could say the public modules on the HX Github aren't designed to be open source (there just isn't anything that is worth hiding in them).

3) What sort of policy should we have about giving commit & publish rights to external people when we are still supporting the modules? We were speaking to an ex-Hxer at nodeconf and he was upset that he had been removed from a module that he had created at HX and was still passionate about. (2) above may help to solve this issue.

shackpank:
@philmeehan Merging these guidelines! Here are other thoughts...

Agree on 1, can we add the holidayextras npm account to those modules as a first step?

On 2, we've had a few chats in group tech and putting everything on the holidayextras github (or leaving them as-is) was the prevailing opinion. Mostly on account of access management getting even more messy if we didn't. The other orgs we look at don't do this - they may have different branding on tech blogs or append "labs" to their name but that's as far as it goes, they have one public github account rather than separating "public" and public.

If an open source module from an existing org (or individual creator) becomes popular enough, it tends to be moved to its own org to reduce the "bus factor", but we have nothing (yet) on the same scale as hapijs, mocha, etc. We should re-evaluate if things go that way!

On 3, it's complicated. Ex-HXers should be clear if they want to own an OSS module before they leave, but should also trust that we'll be reasonable if, for example, GitHub change their permissions model requiring us to re-add whoever to whatever. Talk to us!

PR #65, opened 2015-11-11
pr-template-ip added IP slot in template

Arditc:
# What does this change?

Added IP slot in the Pr-template
# Why?

We would like to allow IP's to be have a slot so that they'll be able to review PR's. This is an optional field which means if you don't get an IP to review the PR it won't block you from pushing your work forward.

PR #68, opened 2015-11-12
Slim down the PR template

dancork:
Following on from the discussion around IPs giving an optional review I'd like to propose an update to the pr template.
### What's changed
- Added whitespace to save time
- Removed wordy dev checklist heading
- Turned checkboxes in to a list
- Added 3 tickboxes to indicate +1
- Removed job titles to make it usable by the entire group

### DIFF
--- a/pr-template.md
+++ b/pr-template.md
@@ -1,20 +1,30 @@
 #### What does this PR do? (please provide any background)
+
+
 #### What tests does this PR have?
+
+
 #### How can this be tested?
+
+
 #### Screenshots / Screencast
+
+
 #### What gif best describes how you feel about this work?
+![](url)
+
 ---
-#### Developer Definition of Done/Quality Checklist (for PR author to complete BEFORE code review):
+
 - [ ] I have checked the [contributing document](../blob/master/CONTRIBUTING.md) and I'm happy for this to be reviewed.
 
-#### Software Engineer or Developer review:
-- [ ] I’ve witnessed the work behaving as expected (this could be on the authors machine or screencast).
-- [ ] I’ve checked for coding anti-patterns.
-- [ ] I’ve checked for appropriate test coverage.
-- [ ] I’ve checked all the tests are passing.
+#### Reviewers (edit comment to add your github username and then +1)
 
-#### Software Engineer or project guru review:
-- [ ] I’ve witnessed the work behaving as expected (this could be on the authors machine or screencast).
-- [ ] I’ve checked for coding anti-patterns.
-- [ ] I’ve checked for appropriate test coverage.
-- [ ] I’ve checked all the tests are passing.
+[ ] IP (optional)
+[ ] Dev
+[ ] SE/Guru
+
+By adding a +1 you are confirming you have...
+- Witnessed the work behaving as expected (this could be on the authors machine or screencast).
+- Checked for coding anti-patterns.
+- Checked for appropriate test coverage.
+- Checked all the tests are passing.
--- a/pr-template.md
+++ b/pr-template.md
@@ -29,7 +29,7 @@
 - [ ] :+1:
 
 By adding a +1 you are confirming you have...
-- Witnessed the work behaving as expected (this could be on the authors machine or screencast).
+- Witnessed the work behaving as expected (this could be on the author's machine or screencast).
 - Checked for coding anti-patterns.
 - Checked for appropriate test coverage.
 - Checked all the tests are passing.
--- a/pr-template.md
+++ b/pr-template.md
@@ -17,15 +17,15 @@
 
 - [ ] I have checked the [contributing document](../blob/master/CONTRIBUTING.md) and I'm happy for this to be reviewed.
 
-#### Reviewers
+#### Reviewers (replace _available with your username)
 
-IP _(optional)_
+IP: _available_
 - [ ] :+1:
 
-Dev
+Dev: _available_
 - [ ] :+1:
 
-SE/Guru
+SE/Guru: _available_
 - [ ] :+1:
 
 By adding a +1 you are confirming you have...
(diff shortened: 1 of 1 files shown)


connormeredith:
I like the idea Dan, definitely streamlines the process :smile:

pmcnr-hx:
I'm not sure about having to edit the PR text to add the reviewer names. Commenting will already add the reviewer as a participant.

dancork:
@pmcnr-hx good point, my only concern there would be that sometimes people comment when not reviewing

Joezo:
I like it! Agree with pedro that we shouldn't edit the PR text. Most of the time people say something along the lines of "dev", "taking se/guru" etc. so it becomes fairly apparent who's reviewing it :)

cameronviner:
Looks good to me, I agree with joe and pedro on the editing the reviewer bit. Also seems a bit weird to me to add a checkbox for an IP if its not a required tick for it to go live, could they not just comment on the PR?

dancork:
Thought adding the name made it easier to see at a glance who was reviewing, the 2nd review can sometimes appear below a lot of comments. I also considered reserving the first/second comments for it. If the consensus is to remove I'm happy to get rid.

@cameronviner I created this PR after the discussions here... https://github.com/holidayextras/culture/pull/65 IPs getting an optional +1 and want a tickbox.

pr-template.md
```diff
@@ -1,20 +1,35 @@
...
+- [ ] :+1:
+
+**SE/Guru**
+- [ ] :+1:
+
+By adding a +1 you are confirming you have...
+- Witnessed the work behaving as expected (this could be on the author's machine or screencast).
+- Checked for coding anti-patterns.
+- Checked for appropriate test coverage.
+- Checked all the tests are passing.
```
kevinhodges, line 35: CI should now be doing this for you, I would think we can drop this.

ghost:
Really like this change, tidies things up and makes it a lot clearer while also implementing the new check for IPs, very nice :)

pr-template.md
dancork, line 35: I think in some instances we can merge if one CI has passed and not another - eg Travis passing but not Team City for Tripapp.

Arditc, line 35: I thought we could too, but new projects in GitHub  might not have CI built in yet.

Arditc, line 35: Also selenium tests are sometimes flaky in some repositoris which means the work could possibly go to production and when someone else works on that repository, they might get builds failing due to the other persons selenium tests.

dancork:
I’ve removed the title to make it potentially usable by the entire group

philmeehan:
:-1: for me if you want SB to use it as well. Sorry :(

I find the checkboxes useful as a reminder to do certain things before submitting the PR, and also when you are code-reviewing the PR.

Think we should look at the other extreme - potentially having a PR template per repo. Some repos will have different things that need to be checked before the PR can be merged. If there is nothing special that needs to be checked then we could use the generic group template, but do think that should have the checkboxes.

JamieMatthews:
@philmeehan is there anyway of having something that can apply to all repos that does not require X amount of check boxes?

I guess the same logic applies to testing instructions on a PR. It's not there to tell us how to do the job, but to point us to where the work would be, then it's down to us to decide what needs to be checked with that particular bit of work. Would this logic apply to code reviewing as well?

If that logic does apply, then that means we could get a streamlined PR process with limited check boxes, making it look tidy and less 'checky'.

**Sorry, seen these PR's popping up so thought I'd share my thoughts :smile:

ghost:
I like this idea,captures everyones requirements, the tick boxes at the top, now being one per reviewer still allows SM's, PO's and ST's to quickly see what stage a PR is in without needing to review all of the comments. 

I like the idea of a reviewer putting there name at the top as it allows the team to see who has ownership over that PR quite quickly and we have reduced the number of tick boxes which makes the process easier. 

@philmeehan I would argue that by removing some of the boxes, we are not removing the check list just moving it as all of the points should be captured in the contributing document.  +1

dancork:
@philmeehan Generally speaking I find when you add extra process/checkboxes it slows everything down. Our PR template was previously about 28 checkboxes which included dev and 3 reviews. Since we slimmed down to 9 the process has been much quicker.

My feeling is that there is nothing stopping yourself and others who see value in lists using them, but we shouldn't enforce it for all. From my experience most people comment, add a :+1: or :shipit: then tick the boxes as a formality.

dancork:
We could trial adding names to the original comment or a reserved first comment

philmeehan:
@dancork: Something that both @hpoom  and @kevinhodges have constantly re-iterated to us is the importance of the code review process - basically the person reviewing the code is more responsible for the code going in than the person who coded it. Based on this, I'm not sure that you would want to speed up the process here.

JamieMatthews:
Adding an extra field for names surely goes against us wanting to slim down the process? It's just replacing the tick box with a name tag, basically. 

Reserving first comment is a good idea +1

I do agree with @philmeehan though, we can't speed this up too much or we will notice an impact on the quality of the PR process, which in turn affects testing and the agile process, to some extent I guess :see_no_evil:

dancork:
@philmeehan by process I mean the bureaucracy rather than the code review. Code review should still be performed in the same way.

dancork:
I get the idea that a list helps a reviewer ensure they've covered everything but I don't see the need for it to be checkboxes. Especially with the 2nd review being performed by a more senior dev, it should be obvious if someone is not reviewing to high enough standard and then that can be fed back to them.

philmeehan:
Sorry I don't want to stop you guys improving your processes so if you feel these changes are going to help you then :+1:. It's more that I don't think SB adopting the same template for all repos will add value at the present time.

We haven't suffered from bureaucracy at this point of the process, and using checkboxes does help us during both the review and the qa process (we use a different template for the 2 stages), but then our whole development process is different to what you guys use.

PR #69, opened 2015-11-13
Moved docs from server to general

hx-markterry:

I've just moved the server side module best practice to the the general JS doc. It has not been changed at all.

Reviews should review this as when we write commonjs on the client the modules should follow the same format as when we write them on the server.

_HX_ 
- [ ] :+1:

_Short breaks_
- [ ] :+1:

### DIFF
--- a/general-javascript-best-practices.md
+++ b/general-javascript-best-practices.md
@@ -145,3 +145,28 @@ var name = "oli";
 var list = [ 1, 4, 5, 7, 9 ];
 doStuff(name, list);
 ```
+
+## Module Template
+
+All modules should follow this order:
+
+1. define a local object and export it
+2. `require()` in any dependencies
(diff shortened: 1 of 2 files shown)


general-javascript-best-practices.md
```diff
@@ -145,3 +145,28 @@ var name = "oli";
...
+2. `require()` in any dependencies
+3. (all the logic goes here)
+
+For example:
+```javascript
+// define a local object and export it
+var ourModuleName = module.exports = { };
+
+// Next, require any dependencies:
+var _ = require('underscore');
```
jamesmcneil, line 163: Although just an example it would be good to use `lodash` here as we prefer it over `underscore` in our projects now?

jamesmcneil:
Makes good sense to me! :+1:

PR #70, opened 2015-11-16
Ditch the checkboxes

robhuzzey:
I am incredibly aware of the irony of using the PR template to review my change to the PR template so I'm not using it.

Simply comment below... no bikeshedding, just do you like this or not :)

### DIFF
--- a/pr-template.md
+++ b/pr-template.md
@@ -7,10 +7,10 @@
 #### How can this be tested?
 
 
-#### Screenshots / Screencast
+#### Screenshots / Screencast (optional) 
 
 
-#### What gif best describes how you feel about this work?
+#### What gif best describes how you feel about this work? (optional) 
 ![](url)
 
 ---
@@ -19,19 +19,10 @@
 
 #### Reviewers
 
-**Review 1**
-- [ ] :+1:
-
-**Review 2** \*
-- [ ] :+1:
-
-**Review 3** _(optional)_
-- [ ] :+1:
+Must be minumum of 2 reviewers consisting of at least one SE, SA or Project Guru.
 
 By adding a +1 you are confirming you have...
 - Witnessed the work behaving as expected (this could be on the author's machine or screencast).
 - Checked for coding anti-patterns.
 - Checked for appropriate test coverage.
 - Checked all the tests are passing.
-
-\*  for HX this review must be completed by an SE, SA or Project Guru


pr-template.md
```diff
@@ -19,19 +19,10 @@
...
 
-**Review 1**
-- [ ] :+1:
-
-**Review 2** \*
-- [ ] :+1:
-
-**Review 3** _(optional)_
-- [ ] :+1:
+Must be minumum of 2 reviewers consisting of at least one SE, SA or Project Guru.
```
DuncanFenning, line 22: We don't have gurus anymore, Guilds are the way forward :)

hpoom:
If we remove the checkboxes, should the questions/what we expect cover off more?

https://quickleft.com/blog/pull-request-templates-make-code-review-easier/

jodiedoubleday:
By removing the checkboxes your replying on looking at people comments to see if the PR is ready to go or now. I worry that if there is a particularly problematic or complex PR then you wont be able to see the thumbs up in all the noise.

We should keep the checkboxes :-1:

robhuzzey:
@jodiedoubleday in practice though, we are still having to chase anyway... checkboxes are sometimes missed.

If we have that many open PR's we need a way to filter them at a glance... we have bigger problems to solve :/

robhuzzey:
@hpoom Yep, that's quite a nice template. Essentially I want the process simplified with no redundant actions. If a lot of those questions are standardised, will we get generic answers with no thought... probably?

I'd even argue, ditch the template totally... each PR is different & should have a good description that is custom for that PR... but I think that's a step too far tight now.

pr-template.md
robhuzzey, line 22: That's a good point... what is a guild person.... a "guildee"?

hpoom:
Personally I like that the checkboxes summarise the acceptance away from the chatter as @jodiedoubleday highlighted. So discussion on the comments, but acceptance all at the top (through the use of checkboxes).

reecemillsom:
Would it be a good idea instead of review 1 etc.. To have dev review, engineer review, IP review?

hpoom:
@reecemillsom See the discussion here https://github.com/holidayextras/culture/pull/68
This has already been discussed at length. Reasons provided were that those roles are not consistent across HX. Also two SEs might choose to that the SE and Dev slot.

pr-template.md
DuncanFenning, line 22: There are Guild Leaders, Guild Members and Guild Associates as far as I remember

reecemillsom:
@hpoom thanks just read through the discussion

hpoom:
I am going to jump in with one more suggestion. It appears the PR template has had lots of discussion recently. This discussion is now split over 2+ PRs. Clearly we are struggling to agree on what PR template should be.

Can I suggest we either need multiple PR templates, which I know people are opposed to, but would appease those that feel the template should differ either by system/repo, or between business streams.

Other option is we put a few people in a room, that this offline and get those people to agree on a single PR template.

I don't mind which but somebody needs to take this own it and get a swift resolution.

robhuzzey:
@hpoom with SB too... I just want it simple, one template for all... no unnecessary tick boxes as well as writing "+1" etc.

Maybe I've opened a can o' worms on this one?

michaelcarter:
The PR template/checkboxes were introduced for a few very valid reasons:
- At a glance checking on PR progress.
- Accountability on the behalf of developers, ticking the box has a
- Making sure all of our PRs are reviewed to a proper standard.

We've moved a lot of the tick boxes that were on the PR template out into the contributing docs for repositories so this template is flexible enough that everyone can use it. The term “bikeshedding” is being thrown around a lot on this - probably fairly! I think everyone would prefer something slightly different to one another where the PR template is concerned, but unless someone's bringing a radical objective improvement to what we already have, I don’t see that the benefits of the PR outweigh the cons of the discussion and disagreement around it.

robhuzzey:
Ok... closing this.... giving the current one a chance to be used :)

PR #73, opened 2015-11-19
S3 deployment info

hx-markterry:
# What does this change?

This documents the outcome of a meeting IT had around how naming our S3 buckets whilst keeping management as simple as possible.
# Why?

Recently we have had several requirements to serve up code from our git repos onto our landing pages.

We have so many S3 buckets at present it is not obvious to IT how each bucket is used.

The new approach will limit the amount of buckets we have (and setup cloudfront and SSL etc), plus allow IT and devs easily see what code goes where.
# Reviews

3 tick boxes one for each team concerned:
- [x] INF
- [x] Distribution Dev
- [x] SB Dev

### DIFF
--- a/deployment-guidelines.md
+++ b/deployment-guidelines.md
@@ -17,3 +17,24 @@
 **Obviously the above is a rough guide and *shouldn't* replace common sense. e.g. If something happens out of hours and a developer can't get hold of a tester, they shouldn't hold off deploying a fix.**
 
 N.B. For expedited issues please refer to the following [guide](expedited-procedure.md).
+
+## Web Assets
+
+In order for new code to be deployed to our customers it first needs to be uploaded to the correct S3 bucket.
+
+A single page or app could use code from multiple S3 locations so to help simplify this we have decided on the following structure:
+
+* One S3 bucket per environment, not per project.
+* Each project has its own directory inside each environment's bucket.
+* Project directories are named the same as the GitHub repository name.
+
+Here is an example of this structure:
+
+    hx-staging-repo
+      hx-tracker
+      hx-wizard
+    hx-production-repo
+      hx-tracker
+      hx-wizard
+
+Please raise an `INF` JIRA if you require a new S3 directory to be created, once actioned you will be given the correct credentials for CI upload your assets.


michaelcarter:
+1

deployment-guidelines.md
```diff
@@ -17,3 +17,24 @@
...
+## Web Assets
+
+In order for new code to be deployed to our customers it first needs to be uploaded to the correct S3 bucket.
+
+A single page or app could use code from multiple S3 locations so to help simplify this we have decided on the following structure:
+
+* One S3 bucket per environment, not per project.
+* Each project has its own directory inside each environment's bucket.
+* Project directories are named the same as the GitHub repository name.
+
```
kevinhodges, line 30: do we want to note here about "versioning" on buckets?  Or at least prompt the conversation.  Backing up is something people don't usually think about when creating this stuff

deployment-guidelines.md
```diff
@@ -17,3 +17,24 @@
...
 
 N.B. For expedited issues please refer to the following [guide](expedited-procedure.md).
+
+## Web Assets
+
+In order for new code to be deployed to our customers it first needs to be uploaded to the correct S3 bucket.
+
+A single page or app could use code from multiple S3 locations so to help simplify this we have decided on the following structure:
+
+* One S3 bucket per environment, not per project.
```
kevinhodges, line 27: Our S3 account is set up the other way with project at the top level.  Does anyone else do it this way?

deployment-guidelines.md
hx-markterry, line 30: I see this as another task as getting consistent naming with our projects is a higher priority i think.

deployment-guidelines.md
hx-markterry, line 27: That was the original idea, but IT agreed that this way was easier to manage now that our repos are growing. Also we can have the staging bucket on IT's staging account and production bucket on the production one.

hpoom:
I assume these naming conventions are only for buckets used for deployment. I assume buckets used for assets are other purposes can use a different naming convention and categorisation?

kevinhodges:
@hx-markterry do we want to standardise file and folder naming at this stage or as a group are we happy to allow people to do whatever they want?

hx-markterry:
@kevinhodges after the environment/repo level i think we dont need any further standardisation at this point. If this becomes a problem we can always introduce some further best practices.

hx-markterry:
@hpoom My main priority was for deployment and aiding IT to manage the number of buckets with consistent security policies. If a project needs to upload some other data to S3 this approach can still be used.

hpoom:
Looks good :+1:

hx-markterry:
@kevinhodges All ok with you?

hx-markterry:
Chatted with INF they are happy

markfermor:
For these asset style buckets - Its really important that there are constraints in place to stop/reduce any chances of things being deleted or moved from one place to another. I think that policy is important to include and/or think about as well. Like in the past if a mistake was made on

kevinhodges:
:+1:

PR #74, opened 2015-11-19
Recommend wrappers for solving Chai linting issues

pmcnr-hx:
Add `dirty-chai` to the list of Chai wrapper package alternatives (since it seems to be the more robust solution handling also plugins like `sinon-chai`) and recommend the use of wrapper packages as the best alternative to the problem.
- [x] Distribution :+1:
- [x] SB :+1:


hx-markterry:
ok with me :candy:

philmeehan:
We tend to just use the ignore-rule directive but do like the thought of some dirty chai :)
:+1:

PR #76, opened 2015-11-27
Created a general contributing doc

hx-markterry:
# What does this change?

This work changes the contributing template in the culture repo to an actual contributing doc and reference it in the PR template.

I have removed / moved some of the content from the old contributing template as we now have specific templates for these.

New projects will now have a minimal contributing doc if at all, which will only state any additional factors to consider when contributing. They will also link back to the general one in this repo.

We don't have an example project setup yet, but this is future work.

I can see new repos having something like this in their contributing doc:

```
Please see our general HX contributing guidelines at .....
In addition to the above please consider:
 * A
 * B
 * C
```
# Why?

As our number of repos grow, having a the same file in every repo detailing how to contribute seems like duplication. 
- [x] Distribution
- [x] Short breaks


CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
+# Contributing to Holiday Extras projects
+
+A BIG thank you from everyone using on of our projects and for taking the time to contribute! :clap:
```
michaelcarter: `one of` instead of `on of`

michaelcarter:
+1 pending typo fix.

philmeehan:
:+1:

PR #77, opened 2015-12-03
Update linting guidelines for using chai

tompricetech:
We've been using `dirty-chai` to prevent linting errors when using chai for our automated tests. This is now in place in trip app and hapi and means that we don't have to add a comment to the top of our files to disable linting. Updated the testing-standard to reflect this.
- [x] :+1:
- [x] :+1::+1:

### DIFF
--- a/testing-standard.md
+++ b/testing-standard.md
@@ -30,10 +30,13 @@ expect(myStub).to.have.been.calledWith('foo');
 
 
 ### Linting Issues
-You may occasionally encounter a slight issue here with our linting, specifically our linting complaining about unused expressions in the tests due to Chai's syntax. There are a couple of viable workarounds for this:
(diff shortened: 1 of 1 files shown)


pmcnr-hx:
I wholeheartedly agree with this! :+1:

GeorgeF72:
:+1:

connormeredith:
:+1:

hx-markterry:
just need someone from sb to +1 this

philmeehan:
I had a play with dirty-chai when I was updating our repos to make-up 6.0 and it did break some of our tests so I think we would prefer to just bypass the linting.

I'll need to have a dig around to see where it was.

philmeehan:
Ok I can't recreate the issue so I guess I did something silly at the time (was trying to get 20 repos updated to make-up at once so was probably rushing).

Personally I would prefer not to have the eslint no-unused-expressions rule at all, but dirty chai doesn't seem all that worse than ignoring the rule so I'll :+1: for myself.

I have sent an email to the other SB developers asking for any issues. If nobody replies back by EOP Tuesday then we shall say it is a :+1: :+1:  from SB :)

philmeehan:
Had 2 :+1: from the other devs in SB and no :-1: , so by my rudimentary maths that is :+1: :+1: :+1:  from us :shipit: :)

PR #80, opened 2015-12-14
React Component Coupling

michaelcarter:

A little reminder on react component coupling, and how to avoid.

---
#### Reviewers

**Review 1**
- [ ] :+1:

**Review 2** *
- [x] :+1:

### DIFF
--- a/working-with-react.md
+++ b/working-with-react.md
@@ -82,9 +82,9 @@ var ContrivedExample = React.createClass({
 
   handleClick() {
     this.setState({ clicked: true });
-	 if (this.props.onClick) {
-	   this.props.onClick();
-	 }
+    if (this.props.onClick) {
+      this.props.onClick();
+    }
   }
 	
   …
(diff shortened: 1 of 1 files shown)


lukehansell:
:+1: looks good to me

working-with-react.md
```diff
@@ -45,9 +45,66 @@ For internationalisation, we're betting on [FormatJS's React Intl](http://format
...
+
+var TrackableContrivedExample = React.createClass({
+  …
+
+  handleClick() {
+    this.props.tracker.track('Tracked a click');
+  },
+	
+  render() {
+    return (<ContrivedExample onClick={this.handleClick} />);
```
dancork: From what I've seen our convention is to use handleClick rather than onClick. I think you'd also lose scope. Might the following be better:

`return (<ContrivedExample handleClick={() => this.handleClick} />);`

lukehansell: not sure about the handleClick naming, but the `() => this.handleClick` is a good point. Alternatively you could use `.bind` to remove the necessity of the anonymous function. I think the AF is our standard though as people don't tend to like bind...

michaelcarter:
Updated @lukehansell, @dancork

dancork:
:+1:

PR #81, opened 2015-12-14
CI best practices - promoting artifacts

shackpank:
Add a note recommending building as few times as possible. Build once, redeploy what you already built to other environments in later steps.

This will speed up the process of getting code to production, and reduce the risk of a non-idempotent build introducing production issues.
- [x] 1
- [ ] 2

### DIFF
--- a/ci-best-practices.md
+++ b/ci-best-practices.md
@@ -63,6 +63,7 @@ Points to note:
 
 ### Do's
 
+* Do create an artifact and promote it toward production (rather than eg re-building the project 3x on integration, staging & production)
 * Do cache any dependency directories (such as `node_modules`).
 * Do fail the CI process as soon as one task fails.
 * Do run the quickest and / or most likely to fail tasks first, to enable a broken build to end quicker.


pmcnr-hx:
:+1:

hx-markterry:
OK with me :house_with_garden:

shackpank:
@holidayextras/shortbreaks Any objections?

hpoom:
@notABluesSinger would love your eyes and :+1: on this if you agree

notABluesSinger:
I completely agree with the intent here, however I would ask what would happen if more than one piece of work gets merged into staging? We should then be re-build and re-test when combined and then deliver to production. How would deployment scripts handle this, or are we categorically not allowing delivery of multiple pieces of work into production in one hit?

hpoom:
Great question @notABluesSinger What are people's thoughts.

Do we have deployment trains with multiple changes on the train? Or do we deploy each piece of work independently?

hx-markterry:
@notABluesSinger if an asset is built in development or staging then it can only be used for promotion if there are no newer commits to the branches we care about. If there are multiple merges into the pipeline only the later asset could be used for promotion as it will also contain the other code changes.

hpoom:
@shackpank This is 4 months old. Chat to @notABluesSinger about his feedback. Then merge or bin this whichever is most appropriate.

shackpank:
I'll come back to this one - agreed it is not as simple as suggested with the CI process we're currently recommending, so just adding this bullet point isn't sufficient to make this our default.

PR #83, opened 2016-01-02
Best Practices for Accessible Web Content

hxmattobee:
This introduces a set of best practices for developing accessible web content with HTML, CSS and JavaScript. It extends and complements the best practices already described for those technologies in their own documents.

### DIFF
--- a/accessibility-best-practices.md
+++ b/accessibility-best-practices.md
@@ -123,8 +123,8 @@ The same link phrase should not be used more than once on the same page, unless 
 
 
 <a name="landmarks"></a>
-## Landmark roles
-Landmark roles are a way of extending the native semantics of HTML elements in order to better describe the structure of a page. They make it easier fo assstive technology users to understand and navigate content.
+### Landmark roles
+Landmark roles are a way of extending the native semantics of HTML elements in order to better describe the structure of a page. They make it easier fo assistive technology users to understand and navigate content.
 
 For a demonstration of landmark roles in action, watch the video on [How ARIA landmark roles help screen reader users](https://www.youtube.com/watch?v=IhWMou12_Vk).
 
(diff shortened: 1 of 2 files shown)


elliottcrush:
Taking review.

elliottcrush:
Really good @hxmattobee. Lots of great content in here :+1:

accessibility-best-practices.md
```diff
@@ -0,0 +1,243 @@
...
+    <label for="year" class="hidden">Year</label>
+    <input id="year" type="text">
+</fieldset>
+```
+
+### Labels for checkboxes and radio buttons
+As well as being explicitly associated with `for` and `id` attributes, labels for checkboxes and radio buttons should appear *after* the control in the markup and be displayed *to the right* of the control on screen. This convention ensures compatibility with the widest range of assistive technologies, and is also consisitent with UI patterns established by Microsoft and Apple.
+
+### Identifying mandatory fields
+A mandatory field should be identified in its label, in a way that doesn't rely on vision, so that it can be announced by assistive technology when the field receives focus. This could be an asterisk (*) or simply say "required". Where an asterisk or other symbol is used, this should be explained at the top of the form (e.g. "fields marked with * are required.") Where all fields in a form are mandatory, it is acceptable to simply say "all fields are mandatory", but probably easier (technically) to adopt the same approach throughout and mark all individual fields.
```
msaspence: Mentioning the use of `<abbr title='required field'>*</abbr>` in required labels might be a nice touch here.

hxmattobee: Good call @msaspence. Added.

jackdclark:
Reviewing :mag:

accessibility-best-practices.md
```diff
@@ -0,0 +1,249 @@
...
+```html
+<label for="location">Flying from?</label>
+<select name="location" id="location">
+    <option value="ABZ">Aberdeen</option>
+    <option value="BFS">Belfast International</option>
+    ...
+</select>
+```
+
+Placeholder text (a default vaule in the field) should **not** be used instead of a separate `label` element. Placeholders are cleared on focus and often not announced at all by screen readers or other assistive technology.
```
jackdclark: ~~vaule~~**value**

jackdclark:
This is great @hxmattobee, nice work! :+1:

Joezo:
Really good @hxmattobee :D

PR #84, opened 2016-01-05
added another prefered lib

hx-markterry:
# What does this change?

I've added a note to prefer `grunt-shell` over `grunt-exec`.
# Why?

After some head banging it seems that `grunt-exec` has a fixed output buffer size, so will fail with an unhelpful error message even if the command it runs succeeds, this is a problem when running on CI.

There is already an open issue on the problem: https://github.com/jharding/grunt-exec/issues/67
- [x] Distribution
- [x] Shortbreaks


pmcnr-hx:
:+1:

hx-markterry:
@holidayextras/shortbreaks anyone want to take a review?

taijuten:
Looking good to me

PR #88, opened 2016-01-12
Add jquery exceptions

hx-markterry:
# What does this change?

This work adds the rule to our global ruleset to allow some jquery exceptions that do not follow the naming convention we would prefer.
# Why?

We had a project specific rule in one of our private projects that was not using our makeup utility.

**Review 1**
- [x] :+1:

**Review 2**
- [x] :+1:


lukehansell:
I agree :+1:

robhuzzey:
happy with this :+1:

PR #95, opened 2016-01-29
Rename title & first iteration on guidelines

carolineerclark:
First pass at pulling best practice together on when/how we raise Jiras across the business. Very open to feedback (especially on my markdown/HTML as this is an area I am learning!) Some areas still to complete - feel free to drop info in from other sources, e.g. Google docs.


raising-jiras.md
```diff
@@ -0,0 +1,87 @@
+#Raising Jiras
+
+<p>Jiras are used at Holiday Extras as a means for teams to communicate what work is required (including data and insights). This doc seeks to outline some general guidelines and best practice when writing a jira.</p>
+
+##Title:
+<p>This is a brief description of the problem to be solved. Be clear and to the point.</p>
+
+**Good example:**
+<p>"Change the endpoints in HAPI to new Horizon service for Flexible Car Hire"</p>
```
robhuzzey, line 9: I would prefer this to not prescribe the "how". Instead, could this be worded "Allow flexible car hire to be sold via our API"?

It's a small adjustment but it makes a difference because the solution may not be to change an endpoint thus sending the developer down a prescribed path.

raising-jiras.md
```diff
@@ -0,0 +1,87 @@
...
+- Developers
+- Insights
+- Testing
+- Requests from the business
+<p>*(Will add in links to headings on 2nd iteration)*</p>
+
+###Developers
+
+**1. Requirements**
+<p>This is the 'what', outlining the details of the work required or the problem to be solved. Some key things to consider:</p>
```
robhuzzey, line 26: Here I think the details of the work should be omitted. In some cases it will be totally necessary for example integrating to a 3rd party API where you must adhere to their procedure but in many cases, a simple "here's what we're trying to solve" is a better approach.

raising-jiras.md
```diff
@@ -0,0 +1,87 @@
...
+<ul>
+<li>Which system does this relate to? (Work across multiple systems to solve the same problem can (should?) be split into separate tickets.</li>
+</ul>
+
+**2. Reason/goal**
```
robhuzzey, line 31: I love this! :)

raising-jiras.md
carolineerclark, line 9: Perhaps this isn't the best example? This jira was raised as a result of conversations that Ollie Buck had with Steve at Flexible, so they had already worked out the 'how/what'. Happy to use a different example for best practice.

I think this flags an interesting point around the knowns/unknowns of a piece of work - you can be much more specific where you have done a lot of background work/investigation already.

raising-jiras.md
carolineerclark, line 26: Could reverse that sentence - start with problem to be solved followed by details of the work. 

At what point does the 'how' become clear? And does this ever get recorded on the jira? Should it?

raising-jiras.md
```diff
@@ -0,0 +1,87 @@
+#Raising Jiras
+
+<p>Jiras are used at Holiday Extras as a means for teams to communicate what work is required (including data and insights). This doc seeks to outline some general guidelines and best practice when writing a jira.</p>
+
+##Title:
+<p>This is a brief description of the problem to be solved. Be clear and to the point.</p>
```
hx-markterry, line 6: You can remove all the `<p>` tags as markdown does not require them.

hx-markterry:
Could we link to this doc from the culture's main table of contents.

raising-jiras.md
```diff
@@ -0,0 +1,87 @@
+#Raising Jiras
```
ghost, line 1: I personally like the simple version when raising a jira:
The WHAT and the WHY - to be outlined by any stakeholder. The WHAT should detail the requirements, idea, bug improvement in what ever way explains this best. 

The WHY should be the business case which will help with prioritisation, the forecasted value, the operational saving on time/cost etc.

I don't think the HOW should be detailed until the work is prioritised. When the work is prioritised the development team will the plan the HOW to build and HOW to test using there own expertise. 

I think there should be a clear distinction between:
1. Raising a jira
2. Definition of Ready

We should be careful we are not spending time planning work that is not going to be prioritised and therefore should make a clear distinction between the two points I raised above. Definition of Ready (DOR) is a tool used by Scrum Teams in order to make sure they have what they need to pull a ticket into a Sprint. The DOR should also be an evolving list and may vary dependant on what each Scrum Team need.

sparkymarky79:
I am going through a change of JIRA practice at the moment too to try and make it easier and clearer to the guys looking at the tickets what is the issue/task and why we are doing it.  I also sometimes Reference previous tickets if they have use.  E.g. We completed something similar before on JIRA OPS-1111 for passing LaunchID parameters through the booking flow into this field in CHIPS.  I didn't get time to do this and really keep that back-log tidy, but I am starting to make time for this stuff.  So for me its all about the 
_what_
_why_
_References_
plus any screen shots, mock-ups and other attachments needed

Maybe I can comment more once I have trailed my new method a couple of sprints

raising-jiras.md
hpoom, line 1: I like a lot of what has been suggested here. I do wonder if we should rename the doc `Raising-Tickets.md` and replace all mention of `Jira` with `ticket. That way these guidelines are universal regardless of which ticketing system we use currently or in future. Also it helps with jargon, jira is not a think, a ticket is a think jira just happens to be the system we use today. Also it would be easier for others outside our organisation to use these guidelines or folk them (remember this is a public open source repo).

raising-jiras.md
hpoom, line 9: I like @robhuzzey suggestion `Allow flexible car hire to be sold via our API` would be a better title. `Change the endpoints in HAPI to new Horizon service for Flexible Car Hire` is too prescriptive and often means some people might do exactly that even if the correct solution is not to change endpoints. Also endpoints probably does not mean much to those outside the dev team and I feel all tickets should be readable by the whole business.

Maybe can have 3-5 examples of great ticket titles, and 3-5 examples of bad ticket titles. I can propose some if that would help?

raising-jiras.md
hpoom, line 31: `split into separate tickets` make we wonder if we should mention chunking up and chunking down in here. It is certainly very relevant to raising tickets.

hpoom:
Overall looks like a good start @carolinewoodhams, I have added some notes where I think we could make it even better.

Does anybody in @holidayextras/shortbreaks have views on this. Being that we also raise tickets and might currently follow different guidelines.

PR #98, opened 2016-02-11
Recommend `catch` when testing promises and asserting in `then`s

shackpank:
This makes debugging failing tests a lot easier.
- [x] Makes sense
- [x] :shipit:


rahulpatel:
Makes sense

pmcnr-hx:
Out of curiosity, have we considered using [`chai-as-promised`](https://www.npmjs.com/package/chai-as-promised)? It seems to make for nicer assertions...

shackpank:
Is already mentioned in `1.1` of this doc @pmcnr-hx :+1:

hxpaul:
This one just caught me out! So :+1:

pmcnr-hx:
Thanks for the clarification @shackpank: :+1:

PR #99, opened 2016-02-22
ES6 - Setting modules true for import/export rules

elliottcrush:
# What does this change

Added `modules: true` for the javascript linting rules. 
## Why

This is another feature of ES6 that i'd like to introduce to the culture linting. A breakdown of what syntax is allowed can be found below:

```

import 'jquery';                        // import a module without any import bindings
import $ from 'jquery';                 // import the default export of a module
import { $ } from 'jquery';             // import a named export of a module
import { $ as jQuery } from 'jquery';   // import a named export to a different name

export var x = 42;                      // export a named variable
export function foo() {};               // export a named function

export default 42;                      // export the default export
export default function foo() {};       // export the default export as a function

export { encrypt };                     // export an existing variable
export { decrypt as dec };              // export a variable as a new name
export { encrypt as en } from 'crypto'; // export an export from another module
export * from 'crypto';                 // export all exports from another module

import * as crypto from 'crypto';    // import an entire module instance object
```

---
- [x] I like this
- [ ] I like this as well


richardstevens:
There are so many positives for this change when it comes to ES6 in the future and currently too.

hx-markterry:
@holidayextras/shortbreaks as this is group tech, could we have a review from you guys too please.

theninj4:
These `export` rules directly contradict our best practices for writing Javascript, which backs up our unit testing best practices. Basically, if we merge this we'll be contradicting how we should be creating modules and how we go about testing those modules. We should patch up those best practices before we let the linting contradict what good code looks like.

We've got a lot of developers now, we need to make sure we're all writing code in the same way or we'll be adding barriers to working well as a team.

elliottcrush:
Would you be able to link to those contradictions directly @theninj4?

hpoom:
@holidayextras/shortbreaks anybody have views on this, being that linting rules affect us all.

dancork:
Personally I'm a big fan of import & export, so for me it's :+1:

theninj4:
Here's two of the headings right at the top of the serverside (less serverside now we're using CommonJS everywhere) javascript best practices:
https://github.com/holidayextras/culture/blob/master/serverside-javascript-best-practices.md#module-template
https://github.com/holidayextras/culture/blob/master/serverside-javascript-best-practices.md#internal-private-functions-should-be-exposed-to-enable-testing

They both set a clear message of "this is how we will export functions to enable testing". When people put in PR's that don't follow the best practices, we can simply link them to the above and ask them to change it, no bike shedding or repeat discussion is necessary.

If we're going to send the message to developers that we're all doing to adopt a new style of exports we need to maintain the same level of quality we're used to, and not forget the learnings from our 4 years of prior javascript experience. I'm sure there's a format of the exports syntax that gives us what we currently have (and probably more) but I'd hate to see us take a step backwards on this one.

kevinhodges:
Can anyone tell me what problem this is solving?  Do we see this as a server or client side solution or both?  Node's require/export syntax already gives us a lot of this functionality.

rahulpatel:
I'd quite like to see us adopt the new import/export syntax, mostly because it'll provide us with _[tree shaking](https://github.com/webpack/webpack/tree/master/examples/harmony-unused)_ in webpack 2 which will result in much smaller assets giving benefit to our customers.

We can still follow our best practices using the new import/export syntax:

_[Example 1](https://github.com/holidayextras/culture/blob/master/serverside-javascript-best-practices.md#module-template)_

``` javascript
// define a local object and export it
export let ourModuleName = { };

// Next, require any dependencies:
import _ from 'underscore';
import moment from 'moment';
import async from 'async';

// Now attach all functionality to our local object:
ourModuleName.somePublicFunction = () => { };
ourModuleName._somePrivateFunction = () => { };
```

_[Example 2](https://github.com/holidayextras/culture/blob/master/serverside-javascript-best-practices.md#internal-private-functions-should-be-exposed-to-enable-testing)_

``` javascript
export let myModule = { };

myModule.mainFunctionality = () => {
  // chain other private functions together
};

myModule._private1 = () => { };
myModule.__private1Part2 = () => { };
myModule._private2 = () => { };
myModule._private3 = () => { };
```

kevinhodges:
I'm not in a position to comment on front end best practice but my gut feel is we don't need this on the back end.

Joezo:
I'm in favour of bringing es6 modules in. The benefits that Rahul mentioned are nice and as said we can easily maintain our existing patterns whilst using them. Enabling them opens up doors for new patterns to  evolve which can only be good for us.

I think one way of solving the "our culture repo has lots of codes examples using commonjs" problem we can add a section to the general-javascript-best-practices document to address modules. I'd like to see something along the lines of "Wherever you see example code within this repo you **can** use es6 modules to achieve the same result." 

It's important to note that node doesn't support ES6 modules at the moment but AFAIK there are plans to adopt them once they've made their way into V8.

elliottcrush:
Apologies for the delay in replying to these comments, is it safe to assume we can develop the existing documentation to differentiate the module declaration between backend and clientside JS? Thereby allowing us to introduce the choice of using ES6 modules?

elliottcrush:
Going to be merging this shortly. After lengthy discussion both in this PR and in the front-end guild chat last week we've decided that we want this feature on.

msaspence:
@elliottcrush can we get references to other PRs on the culture repo or on makeup repo as they appear

richardstevens:
@elliottcrush any movement on this do you know? 

We are blocked on the static-site-generator repo using make-up

theninj4:
@richardstevens it looks like we're all moving away from custom linting rules and make-up and slowly moving towards using `standard` https://standardjs.com/ instead.

richardstevens:
Thanks @theninj4 ill install that instead of make-up and give it a go 👍

elliottcrush:
Agreed @theninj4 and on that note. I will close this PR

PR #105, opened 2016-03-11
Update .eslintrc file to be compatible with eslint 2

Joezo:
Updates .eslintrc file to be eslint 2.x compatible. 
This removes the individual ES6 feature flags. I can't see a way of disabling certain parts anymore. It makes sense, you can't disable individual parts of ES5. It's important to note that modules are still disabled with this config, you have to explicitly turn these on: http://eslint.org/docs/user-guide/migrating-to-2.0.0#language-options
- [x] :+1: 
- [x] :+1: 
- [x] :+1:

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,36 +9,18 @@
     "amd": true,
     "browser": true,
     "jasmine": true,
-    "jquery": true
+    "jquery": true,
+    "es6": true
   },
   "plugins": [
     "mocha",
     "react"
   ],
-  "ecmaFeatures": {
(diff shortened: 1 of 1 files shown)


notABluesSinger:
Taking Slot :one: 
:+1:

hxpaul:
This works for me, tried it out on a project updating eslint and eslint-plugin-mocha to v2 and using this rule set and it still looks sensible :+1:

pmcnr-hx:
Good job: :+1:

PR #107, opened 2016-03-21
[WIP] First draft of new testing guidelines

michaelcarter:
**Please read:**
This is part of the testing guild's effort to update our testing guidelines, and push the concept of integration tests more as a viable test type to use here at Hx.

I've written this as a separate document for now so we can get a discussion going around just what's in here, but before merging I'd like to integrate the current testing standard into this document, probably dropping some of the stricter "rules" along the way in favour of more flexible guidelines that allow for the range of systems we now have in place here.

homer bird hair

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -47,9 +47,11 @@ Integration tests are particularly useful for:
 However, integration tests have some drawbacks:
  - Expensive to write
  - Expensive to maintain
+ - Only accessible to developers
  - Dealing with real objects means that test setups can be hard to debug. Not everything is stubbed so finding what's broken a test, checking whether it should or should be stubbed and then addressing the issue can be a time consuming process.
  - They can test the code works as expected, but don't prove the application actually makes sense to the user.
  - Can be prone to slowdowns when left unchecked.
+ - Integration tests at the low-level extreme are sometimes better written as unit tests, and at the high-level extreme are sometimes better written as system tests. They can be a tricky judgement call to make sometimes.
 
 ### System Testing
 [System Testing](https://en.wikipedia.org/wiki/System_testing) involves testing the system as a whole. It sits "above" unit and integration testing conceptually and tests the application by using it the way the end user would use it. The goal with system tests is to make sure the application works as a product and that that it can be shown to perform its business requirements. Taking making a cake as an analogy, you can think of system testing as checking the baked, and iced carrot cake tastes like a good carrot cake.
@@ -58,13 +60,12 @@ System tests are particularly useful for:
  - Testing user journeys throughout the application. Proving the application meets business requirements.
  - Making _some_ testing available to non-developers. _Some_ system testing frameworks are human friendly enough that anyone can learn to write good system tests in a reasonable amount of time. This applies mainly to applications with a user interface.
  - Testing very complex regressions that are not practical to unit or integration test.
- - Testing visuals on applications with user interfaces. Some system testing frameworks allow us to generate, store and compare screenshots as well as testing user interface for clickability.
+ - Testing visuals on applications with user interfaces. Some system testing frameworks allow us to generate, store and compare screenshots as well as testing user interface elements for clickability.
 
 However, system tests have some drawbacks:
  - Slow to run. This can be overcome with parallelism, but that can be expensive and the suite will only ever be as fast as the slowest test.
  - Brittle and prone to false negatives. This can be overcome with data fixtures and awareness of best practices.
- - Loosely constrained. In laymens terms, your application code could be doing some horrible stuff under the hood, but your system test is only concerned with inputs and outputs at the system level.
-
+ - Loosely constrained. Your application code could be doing some _horrible_ stuff under the hood, but your system test is only concerned with inputs and outputs at the system level and so that probably won't be exposed in your system test.
 ---
 
 Different systems will be suited to a different mix of the above test types. Choose wisely based on what it is you're testing.
--- a/testing.md
+++ b/testing.md
@@ -1,21 +1,21 @@
 # Testing
 
 ## Why write tests?
-
-We write automated tests to allow us to quickly demonstrate that our software meeting business requirements. They're one of a few tools we use to increase our confidence in our work prior to deployment.
+We write automated tests to allow us to demonstrate that our software meets business requirements, and that it continues to do so as changes are made to the codebase.
 
 ## When do I write tests?
-
-### Test necessity
-_"If this test fails, what business requirement is compromised?"_
-
-This is a good question to ask yourself before writing tests. If the answer is _"None"_, you shouldn't be writing it. If the answer is _"I don't know"_, it's possible the test you're writing has no value.
+Almost all of our code should be tested to some degree, the question is less about "when" and more "to what level do I test this code?". There are a few points to consider to help with this decision making.
 
 ### Net costs
-It's also worth considering the net cost of tests before you write them. For example, if the business impact of a regression is small and unlikely enough, it could be that you're probably incurring a net cost to the business by writing a test that needs to be **1.** understood, **2.** run as part of your CI and **3.** maintained.
+It's worth considering the net cost of tests before you write them. Simply put, ask yourself "Is this test likely to pay for itself over the course of it's lifetime?".
+
+If not, it could be that you're incurring a net cost to the business with a over-thorough test that needs to be **1.** understood, **2.** run as part of your CI and **3.** maintained. In this scenario it might be worth opting for a something less expensive that still gives the team an acceptable level of confidence that the code works as expected.
 
 ### Keep the system in mind
-Lastly, when making decisions about the scope of your testing, keep the system in mind that you're working on. For example, a single failure on a data-sensitive area of the API is potentially going to affect hundreds of customers and could expose masses of customer data. Whereas a single failure in React component is only going to affect a single session at once, and probably won't be a showstopper.
+When making decisions about the scope of your testing, keep the system in mind that you're working on. For example:
+ - A single failure on a data-sensitive area of an API is potentially going to affect hundreds of customers and could expose customer data. This needs extensive, thorough, testing to ensure customer data is never exposed.
+ - Conversely, s single failure in a simple React component is only going to affect a single session at once, and might just incur a slight visual glitch or formatting issue. We probably want to write a quick test to check this works as expected, but testing every possible edge case is likely to cost us more than the test is worth.
+ - Library code should always be tested extensively It's designed to be reused many times across many systems in many use cases - we need to be confident this works as advertised no matter how it's used.
 
 ## Types of Testing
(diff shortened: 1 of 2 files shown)


testing.md
```diff
@@ -0,0 +1,71 @@
+# Testing
+
+## Why write tests?
+
+We write automated tests to allow us to quickly demonstrate that our software meeting business requirements. They're one of a few tools we use to increase our confidence in our work prior to deployment.
```
theninj4: `s/meeting/meets/g`

theninj4: I disagree with this statement, we don't write tests to prove our code does what we say it does. We write tests to enable other developers to contribute to our codebases without introducing regressions, in other words "we write tests to stop other people breaking our stuff". If we ever decide to write tests to prove developers can develop, we're basically saying we don't trust people to do their job.

testing.md
```diff
@@ -0,0 +1,71 @@
...
+
+## When do I write tests?
+
+### Test necessity
+_"If this test fails, what business requirement is compromised?"_
+
+This is a good question to ask yourself before writing tests. If the answer is _"None"_, you shouldn't be writing it. If the answer is _"I don't know"_, it's possible the test you're writing has no value.
```
theninj4: This is somewhat vague and open to interpretation. If I can unintentionally break your feature and no tests fail, then it needs better tests.

testing.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Net costs
+It's also worth considering the net cost of tests before you write them. For example, if the business impact of a regression is small and unlikely enough, it could be that you're probably incurring a net cost to the business by writing a test that needs to be **1.** understood, **2.** run as part of your CI and **3.** maintained.
```
theninj4: Historically I've seen a lot of tests very tightly coupled to the code they're testing. As in, if I were to change the code in a mediumly significant way, I'd have to spend a LOT of time trying to piece together all of the failing tests it introduces. If we write our tests in the right way, the cost of maintenance drops towards the point at which it becomes negligible. In the ideal world, test maintenance is a non-issue.

testing.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Keep the system in mind
+Lastly, when making decisions about the scope of your testing, keep the system in mind that you're working on. For example, a single failure on a data-sensitive area of the API is potentially going to affect hundreds of customers and could expose masses of customer data. Whereas a single failure in React component is only going to affect a single session at once, and probably won't be a showstopper.
```
theninj4: Are we saying that we'll accept lower quality tests on projects that have less customer impact? Test quality and code quality go hand in hand - good tests demand good code. If we allow substandard tests, we're encouraging substandard code.

testing.md
```diff
@@ -0,0 +1,71 @@
...
+ - Proving the functionality of complex algorithmic logic.
+ - Alerting developers to potential regressions during refactors in a "yes, I really did mean to do that there" kind of fashion.
+ - Providing 100% thorough exercising of business critical code. E.g. Ensure a card number validator does not block valid cards.
+ - Testing potential edge cases for functions. E.g. "What happens when I throw garbage into this?"
+
+However, unit tests have some drawbacks:
+ - Expensive to write
+ - Expensive to maintain
+ - Only accessible to developers
+ - Hard to determine a business purpose from. E.g. "I can see this code is doing this and that, but what does that actually _mean_ in product terms?"
```
theninj4, line 36: I somewhat disagree with the points above, and the point below I see as being more of a definition of a unit test rather than a drawback. It feels like these three blocks are framed somewhat negatively around writing tests all together?

Maybe we should gather a bunch of people together and talk through the different types of tests, open it up to some discussion?

testing.md
michaelcarter: I assume your disagreement is with the, "confidence in our work prior to deployment" rather than the sentence before? When I wrote that, I meant "confidence in our work" as "confidence in our software meeting business requirements now and in the future", not confidence that our code does what we want it to. Business requirements encompasses regressions, speed, security, functionality etc?

testing.md
michaelcarter: What do you mean by "negligible" and "non-issue" here? Do we have examples of systems here or elsewhere in which demonstrate test maintenance as a non-issue or with a negligible cost?

testing.md
michaelcarter: I think the test code itself should be of a consistent quality, but the thoroughness of the tests should be appropriate to what it is we're testing. Again I think this comes back to cost to write and maintain vs. lifetime value to business.

testing.md
michaelcarter, line 36: I included all of these points because people tend to write either unit or system tests at Hx. By laying down all the pros and cons of them from an overall quality assurance perspective, I'm hoping people are able to make informed choices themselves about what's best for their code. I agree that the last point is more the point of unit testing than an inherent disadvantage though.

testing.md
theninj4: It was actually with the first statement, it reads like we're saying "we write tests to prove our code does what we think it does".

testing.md
theninj4: We definitely have old tests kicking around in some closed-source projects whereby changing the behaviour of the code in a minor way would take an exponential amount of effort to fix up the tests. On the flip side, we also have a lot of tests whereby we only need to update a couple of lines of tests to reflect some new changes and get the build passing.

theninj4: `Negligible` as in, only need to update a handful of lines of code.
`Non-issue` as in, making the changes is so trivial that avoiding it would be... lazy?

testing.md
msaspence: Whilst I don't disagree with the statement "we write tests to stop other people breaking our stuff" do I disagree with the idea that "we don't write tests to prove our code does what we say it does". Writing tests like this doesn't say "we don't trust people to do their job", it says "we understand that people are human and make mistakes", automated testing is a way to catch some of these mistakes.

I don't think there is really a huge difference between these two ways of thinking about it though. If you manage to break the application... then it no longer behaves as expected. Both are a valuable ways of thinking about how to write tests.

In any event I think the latest version reflects both of these.

testing.md
msaspence: "Quality" is fairly vague and subjective on its own. Time to write, time to read, time to run, time to maintain, false negatives blocking builds, coverage, edge cases, potential impact of false positive, etc all contribute to the value and cost of the test. Its not about "lower quality tests" its about weighing the cost and value of each test. A test that appropriately accounts for the above considerations is in my book a "high quality" test.

testing.md
msaspence, line 36: I think the last point is valid, your right @theninj4 it is the point of unit tests, but when introducing these concepts to perhaps more junior developers it is an important thing for them to keep in mind.

testing.md
```diff
@@ -0,0 +1,72 @@
...
+ - Alerting developers to potential regressions during refactors in a "yes, I really did mean to do that there" kind of fashion.
+ - Providing 100% thorough exercising of business critical code. E.g. Ensure a card number validator does not block valid cards.
+ - Testing potential edge cases for functions. E.g. "What happens when I throw garbage into this?"
+
+However, unit tests have some limitations:
+ - Expensive to write
+ - Expensive to maintain
+ - Only accessible to developers
+ - Hard to determine a business purpose from. E.g. "I can see this code is doing this and that, but what does that actually _mean_ in product terms?"
+ - Provide no guarantee that when all of your code is put together, that anything actually works as the business expects.
```
msaspence, line 37: - Hard to write upfront if you are following a TDD methodology

testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Types of Testing
+There are three major categories of testing that might be appropriate to use in your development.
+
+### Unit Testing
+[Unit testing](https://en.wikipedia.org/wiki/Unit_testing) involves testing code at a very granular level, usually testing a single function and everything that goes on within it, stubbing calls to other functions or modules to ensure we can accurately check the functionality of the unit. Taking making a cake as an analogy, you can think of unit testing as checking the quality of the individual ingredients.
+
+Unit tests are particularly useful for things like:
+ - Proving the functionality of complex algorithmic logic.
+ - Alerting developers to potential regressions during refactors in a "yes, I really did mean to do that there" kind of fashion.
```
msaspence, line 28: Unit tests only offer value for refactor if the refactor is within the unit. If for example you are refactoring a module and you've tested at the function level you will likely need to either throw out your tests or substantial rewrite them.

testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+However, integration tests have some limitations:
+ - Expensive to write
+ - Expensive to maintain
+ - Only accessible to developers
+ - Dealing with real objects means that test setups can be hard to debug. Not everything is stubbed so finding what's broken a test, checking whether it should or should be stubbed and then addressing the issue can be a time consuming process.
+ - They can test the code works as expected, but don't prove the application actually makes sense to the user.
+ - Can be prone to slowdowns when left unchecked.
+ - Integration tests at the low-level extreme are sometimes better written as unit tests, and at the high-level extreme are sometimes better written as system tests. They can be a tricky judgement call to make sometimes.
+
```
msaspence, line 55: - Can be difficult to cover all the necessary edge cases

testing.md
```diff
@@ -0,0 +1,72 @@
...
+System tests are particularly useful for:
+ - Testing user journeys throughout the application. Proving the application meets business requirements.
+ - Making _some_ testing available to non-developers. _Some_ system testing frameworks are human friendly enough that anyone can learn to write good system tests in a reasonable amount of time. This applies mainly to applications with a user interface.
+ - Testing very complex regressions that are not practical to unit or integration test.
+ - Testing visuals on applications with user interfaces. Some system testing frameworks allow us to generate, store and compare screenshots as well as testing user interface elements for clickability.
+
+However, system tests have some limitations:
+ - Slow to run. This can be overcome with parallelism, but that can be expensive and the suite will only ever be as fast as the slowest test.
```
msaspence, line 66: "the slowest test and the cost to initialise the test suite"?

testing.md
```diff
@@ -0,0 +1,72 @@
...
+ - Brittle and prone to false negatives. This can be overcome with data fixtures and awareness of best practices.
+ - Loosely constrained. Your application code could be doing some _horrible_ stuff under the hood, but your system test is only concerned with inputs and outputs at the system level and so that probably won't be exposed in your system test.
```
msaspence, line 68: This can be a good thing, if you have a horrible system that you are stuck with for whatever practical reason.

msaspence:
Worth mentioning regression tests?

hpoom:
@holidayextras/shortbreaks any thoughts on this PR?

michaelcarter:
@hpoom We (matt spence, oli r, steve coppin-smith and I) had a meeting around this a couple of weeks ago. We really need to either pick it back up at some point or close it. I'll arrange something this week.

PR #108, opened 2016-03-22
Remote working etiquette

steve-at-hx:
Guidance on best practice for working remotely.

### DIFF
--- a/remote-working-etiquette.md
+++ b/remote-working-etiquette.md
@@ -6,10 +6,14 @@ Working remotely is a great way to get focussed, balance work and life, or maybe
 
 There is no list of acceptable reasons for when it's "ok" to work remotely. You'll be the best judge of whether not being physically present/absent is going to positively or negatively impact the success of the team. Here are some questions you might want to factor in to that decision:
 
+Environment
 - Will I be able to have good virtual access to the team from my remote location (i.e. internet speed able to support a Google Hangout)?
 - Will my environment allow me to focus on my work (i.e. minimal distraction from media, people and of course pets)?
 - Will my environment allow me to communicate easily with colleagues (i.e. minimal background noise and the ability to speak clearly/openly)?
+
+Commitment
 - Will you still be able to make your commitment to any meetings?
+- Will working remotely hinder your ability to hit your sprint commitments?
 
 We all have various degrees of "cost" to being in the office, especially as we grow the team wider. It may be that you can peak personal productivity by saving your commute, immersing in the zen of working from the garden, being home to welcome a package so you don't have the chase around depots in the evening...
 
--- a/remote-working-etiquette.md
+++ b/remote-working-etiquette.md
@@ -1,0 +1,25 @@
+# Remote working etiquette
+
+Working remotely is a great way to get focussed, balance work and life, or maybe just have a change of scenery and be inspired. However, it's really important that while we enjoy these benefits and seek our own state of super-productivity, we remain conscious of the effectiveness of our colleagues and the team around us 
+
+## Remote considerations
+
+There is no list of acceptable reasons for when it's "ok" to work remotely. You'll be the best judge of whether not being physically present/absent is going to positively or negatively impact the success of the team. Here are some questions you might want to factor in to that decision:
+
+- Will I be able to have good virtual access to the team from my remote location (i.e. internet speed able to support a Google Hangout)?
+- Will my environment allow me to focus on my work (i.e. minimal distraction from media, people and of course pets)?
+- Will my environment allow me to communicate easily with colleagues (i.e. minimal background noise and the ability to speak clearly)?
+- Will you still be able to make your commitment to any meetings?
+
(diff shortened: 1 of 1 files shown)


steve-at-hx:
Very early draft and thoughts are my own, so needs lots of reviewing and input from others!

DuncanFenning:
I like this! It makes us ask questions over remote working rather than just doing it. I think if this doc will be completely covering remote working etiquette, we should elaborate over communication. This could be to do with letting people know you intend to work remotely and communicating the work you are doing during the day (necessary updates etc). Still, thinking about possible implications of working remotely is something we don't really do, which does impact the work we do.

steve-at-hx:
> This could be to do with letting people know you intend to work remotely and communicating the work you are doing during the day (necessary updates etc). 

Per our Slack convo, I'm expecting in the medium term we'll need to have pods set up to facilitate remotes to accommodate Sofia devs. Once we're set up it _shouldn't_ have an additional impact if you're not physically present because the pod is set up to accommodate you already, so the need to give notice is less. Maybe giving notice of working from home is short term etiquette that should be observed and therefore added to this. 

I think we should also add something about being as communicative remotely as you would be in the office.

sam-cook:
Maybe one point ‘Am I confident that I will be able to run all systems (i.e VM) from the location I will be working from’ or something similar?

remote-working-etiquette.md
```diff
@@ -0,0 +1,30 @@
...
+- Will you still be able to make your commitment to any meetings?
+- Will working remotely hinder your ability to hit your sprint commitments?
+
+We all have various degrees of "cost" to being in the office, especially as we grow the team wider. It may be that you can peak personal productivity by saving your commute, immersing in the zen of working from the garden, being home to welcome a package so you don't have the chase around depots in the evening...
+
+## The key question
+
+Once these questions have been answered, you'll need to ask yourself: 
+
+Is the **benefit to the team** of me working remotely *greater than* the **impact on the team** of not being physically present.
```
lukehansell, line 25: I'd suggest the benefit here is probably personal rather than to the team?

steve-at-hx, line 25: I agree, Luke, but that has a knock on benefit to the team. If you're personally trying to save a commute, that allows you to either spend more time or be more energised on team stuff. If you're waiting in for a package because you'll be happier knowing it's taken in and therefore more focussed on team stuff. 

Considering the benefit to the team makes it a balanced "equation", and what it's trying to avoid is a decision where it's great for me, not good for the team, as that will likely cause issues.

Interested in your thoughts on this.

msaspence, line 25: There are lots of personal benefits (and therefore indirect business benefits).

But there are also direct business benefits. Essentially the sentiment that (at least for some people) offices are awful places to works. They are noisy, distraction filled rooms that can sap productivity.

jamesmcneil, line 25: @steve-at-hx Although I'm sure it's not the intention, to me the overall balance of this _feels_ weighted more toward suggesting working in the core office rather than being flexible about working locations and patterns. I think most of the  points you are trying to get across are already covered by the company-wide Anytime+ Google doc on my.hx? The two key lines therein for me are `Essentially, if you can be connected and contactable, Why does it matter where you do your work?` and `HX trusts you... lets keep it that way`, I think we could do with incorporating these in there somewhere or maybe we've duplicated enough that we don't need a remote working doc in this repo?

cholness, line 25: I think rather than suggesting that working in core office is by default a 'better' option the intention is to suggest benefits and disadvantages of either working location in a particular instance when we each plan our time in advance. This definitely should not detract from the Anytime+ policy on my.hx but could help us to consider at which times working solo remotely could present the most benefits and when being co-located could offer the most benefits.

steve-at-hx, line 25: Thanks all.

@jamesmcneil as discussed on our call, I think you've surfaced an interesting point about the Anytime+ document. I think I would challenge that it's as simple as the doc lays out and there is another factor there regarding the cost vs benefit of getting in to the office. The example we discussed is whether we'd be completely happy if someone who worked a short walk from The Wave worked from home 5 days a week... However, if we wouldn't be happy with this and would therefore slightly contradict the Anytime+ doc that's a wider issue. Will follow up with @lizpope for her thoughts.

remote-working-etiquette.md
```diff
@@ -0,0 +1,30 @@
...
+Environment
+- Will I be able to have good virtual access to the team from my remote location (i.e. internet speed able to support a Google Hangout)?
+- Will my environment allow me to focus on my work (i.e. minimal distraction from media, people and of course pets)?
+- Will my environment allow me to communicate easily with colleagues (i.e. minimal background noise and the ability to speak clearly/openly)?
+- Will I be able to run all systems (i.e VM) necessary to complete my work and support the team?
+
+Commitment
+- Will you still be able to make your commitment to any meetings?
+- Will working remotely hinder your ability to hit your sprint commitments?
+
```
slavcho-slavchev-hx, line 18: Just a small observation for using first person narrative in the Environment section and second person narrative in the Commitment section. Probably sticking to one would be better?

PR #112, opened 2016-04-01
adding to lodash best practices

lukehansell:
Adding examples to the lodash best practice based on discussions in the front-end guild.
- [x] +1
- [ ] +2

### DIFF
--- a/general-javascript-best-practices.md
+++ b/general-javascript-best-practices.md
@@ -149,6 +149,27 @@ Use instead of underscore, due to correct semantic versioning, better performanc
 
 Please only require the methods you need rather than the whole library in order to keep build sizes at a minimum.
 
+**Good:**
+```javascript
+const _ = {
+  includes: require('lodash/collection/includes')
+};
+```
+
+**Bad:**
+```javascript
+const _ = require('lodash');
+```
+
+As an alternative (and where Lodash is not currently a dependency) you could install and require each Lodash package individually:
+
+**Better:**
+```javascript
+const _ = {
+  includes: require('lodash.includes')
+};
+```
+
 ### grunt-shell
 
 Use instead of grunt-exec, due to a configurable max output buffer size option. Required if running a command with a lot of output.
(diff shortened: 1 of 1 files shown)


general-javascript-best-practices.md
```diff
@@ -149,6 +149,27 @@ Use instead of underscore, due to correct semantic versioning, better performanc
...
+```
+
+**Bad:**
+```javascript
+const _ = require('lodash');
+```
+
+As an alternative (and where Lodash is not currently a dependency) you could install and require each Lodash package individually:
+
+**Better:**
```
Joezo: I disagree with this being the "better" option. I think this is only better if you're using a version of lodash that doesn't let you `require('lodash/func')`. In v4 for example you can just do `require('lodash/includes')`. There is no point adding extra multiple dependencies when we can just install it all in one go. 

I think we should always use the _good_ option. It's probably worth adding that when using the _good_ option you can't use `_.chain`. Or at least i'm yet to find a way of using it in that way.

lukehansell: based on this and other feedback I'll remove the `better` option

hx-markterry:
ok with me :bank:

general-javascript-best-practices.md
```diff
@@ -149,6 +149,18 @@ Use instead of underscore, due to correct semantic versioning, better performanc
 
 Please only require the methods you need rather than the whole library in order to keep build sizes at a minimum.
 
+**Good:**
```
Joezo, line 152: I think we should prefer that people don't use `_.chain` in clientside code. There is no way of using that with this approach. We also lose all the benefits of doing it this way if someone requires the whole of lodash just to use `_.chain`. Thoughts?

lukehansell, line 152: I agree if it's not possible to use `_.chain` in this way but all other methods can still be required. I'll merge this and then we can add another PR to discuss and advise on `_.chain`

lukehansell:
any further objections to me merging this?

Joezo:
not from me :)

PR #118, opened 2016-04-25
Testing overview

JackMoggy:
# What is this document?

This is a document already in use as a process:
We introduced a new approach to giving a Software Testing overview to new members of the Web Team. 

Each new starter will spend a couple of hours with a Software Tester and be given an overview of testing. 

This does not apply to new starters that are Testers as they will need to sit with a Tester from each pod and learn about each environment and testing requirements in more detail.


JackMoggy:
Merging 👍

PR #119, opened 2016-04-27
added paragraph for pairing with tight deadlines

hx-damian-casey:
Added a short paragraph to the pair programming section. 

Check the section under the heading 'Pairing with a tight deadline'


hx-damian-casey:
The new section to check is the `## Pairing with a tight deadline` section.

PR #121, opened 2016-05-04
Pod remote working best practices

cholness:
This is an early draft and thoughts are mostly my own, Please feel free to review and add your input

### DIFF
--- a/pod-remote-working.md
+++ b/pod-remote-working.md
@@ -1,0 +1,37 @@
+## Pod Remote Working - Recommended Practices
+
+### General
+
+#### Video Conferencing (Hangouts/Slack)
+
+* Check meeting room is equipped with good quality camera and microphone.
+* Pod members can mute their microphone when not speaking to improve audio quality for other participants.
+* Quality of remote network connection can impact video/audio conferencing ability. Try audio only if video is not possible.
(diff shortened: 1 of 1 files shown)


steve-at-hx:
I’d like to see that post really positioned as what a pod can do to better facilitate remote working. It seems a little bit of a mix of stuff at the moment. I think I fed back before that it’s lacking a bit of a narrative. Could do with an intro para on why pods should work hard to facilitate remote working and why it’s worth investing in, then perhaps the rest becomes an FAQ of potential issues and how they can be overcome?

msaspence:
I think the language in this could do with being a bit more strongly opinionated. "Cans", "coulds" and "Positive feedback received" are a bit "use this if you feel like it", which has it place but to a point and I think strong leadership also has value.

I appreciate its a first draft and hasn't had much input yet but just because it is strongly opinionated doesn't mean that it can't be subject to a lot of change. I find that more assertive language with these things is more likely to provoke feedback and as long as people are open to and prepared to incorporate that feedback that is a good thing. 

Just my opinion though, interested what other people's thoughts are as this idea has implications way beyond this PR.

cholness:
Thanks both. I have added some more narrative and I agree with you Spence that the language could do with some more firming up. 
I'll make some more tweaks. Appreciate both your input.

PR #125, opened 2016-06-03
Delete .DS_Store

hpoom:
Deletion of .DS_Store, should not be in here. Will add a .gitignore rule to keep it out.


pmcnr-hx:
:+1:

PR #127, opened 2016-06-05
`.eslintrc`: no multiple empty lines

pmcnr-hx:
Add `no-multiple-empty-lines` to our `.eslintrc`.

This rule enforces a maximum of 2 empty lines and a maximum of one at the end of file (catching some files with multiple empty lines at the end we have floating around). This is purely cosmetic but nice to have and can be automatically fixed.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -88,6 +88,7 @@
     "no-mixed-spaces-and-tabs": [2, false],
     "no-multi-spaces": 2,
     "no-multi-str": 2,
+    "no-multiple-empty-lines": [2, { "max": 2, "maxEOF": 1 }],
     "no-native-reassign": 2,
     "no-negated-in-lhs": 2,
     "no-nested-ternary": 2,


ghost:
@pmcnr-hx, thanks for your PR! By analyzing the blame information on this pull request, we identified @Joezo, @michaelcarter and @jackdcrawford to be potential reviewers

rahulpatel:
Why 2?

pmcnr-hx:
Why not? I just wanted the `maxEOF` bit to be honest, but the rule breaks without the `max` option, so it needs to be there. More than `2` empty lines seems to be wasteful. Would you prefer `3`?

rahulpatel:
Ah interesting. I'd say go the other way and limit it to `1`, but i'm happy with whichever number we go with :). 👍

Nick-Patrick:
I'd be happy with 1. Can't think of a case where 2 would be preferred.

pmcnr-hx:
I like 2 to separate requires from the rest of the code for example. There are some exceptions where 2 spaces may be useful. I think it would be too strict a constraint just the 1 space. Also remember this applies to all our projects which might have a different opinion in this regard.

rahulpatel:
Oh I see, I thought this was the rule for `BOF`. `2` seems fine in that case.

pmcnr-hx:
So to clarify: 2 lines in the file, 1 at the end of file.

Nick-Patrick:
I'm happy with it.

hxpaul:
I like this 👍

hx-markterry:
ok with me

PR #130, opened 2016-06-21
Automated Testing Commandments

theninj4:
In an attempt to enable creative thinking without compromising what good testing looks like, I present the 7 commandments of automated testing.


theninj4:
@msaspence, @michaelcarter, @kevinhodges I'd appreciate some feedback on this one

testing-commandments.md
```diff
@@ -0,0 +1,22 @@
+# HolidayExtras Automated Testing Commandments
+
+## Thou shalt be consistent within projects
+If a project has a de-facto standard, it should be maintained. The process for changing a de-facto standard should take the form of a whiteboard meeting, not a pull request.
```
msaspence: I think the "road map standard" should be the go to consideration where it exists, in a perfect world this will also be the de facto standard. For example on the trip planner we have a lot of unit tests, but the road map is to continue to use those where appropriate, but add more integration and system level tests.

"The process for changing a de-facto standard should take the form of a whiteboard meeting, not a pull request." I agree with the value of such a meeting, but I think a pull request, or pull requests are an important part of the process. 1. Because it gives visibility to those who were not in the white board session. 2. There might be details that come up in PRs that were missed or glossed over in any upfront meeting.

Potentially each project should have its own testing guidelines/info on current state/road map that expands on this one.

testing-commandments.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Thy tests should be obvious
+If the test behaviour isn't obvious by looking at it, it's no good.
+
+## Thy tests should be loosely coupled with thy code
+If I make a single code change and many tests fail, it's no good.
```
msaspence: "code" is a bit ambiguous I wonder if we can make this a bit clearer by saying it is ok for tests to be couple to the interface but not to the implementation.

If you are changing the thing you are testing (the interface) you are always going to have to change the tests by definition.

testing-commandments.md
```diff
@@ -0,0 +1,22 @@
+# HolidayExtras Automated Testing Commandments
+
+## Thou shalt be consistent within projects
+If a project has a de-facto standard, it should be maintained. The process for changing a de-facto standard should take the form of a whiteboard meeting, not a pull request.
+
+## Thy tests should be obvious
+If the test behaviour isn't obvious by looking at it, it's no good. If I have to scroll for days to read a test, it's no good.
+
+## Thy tests should be loosely coupled with thy code
```
msaspence: "Code" feels ambiguous, I wonder if we should clarify that tests can be coupled to the code interface (what they are testing) but should avoid coupling to the implementation.

testing-commandments.md
```diff
@@ -0,0 +1,22 @@
+# HolidayExtras Automated Testing Commandments
```
msaspence: Commandments is a fun analogy but there are always exceptions to rules and there may be occasions where the below can be broken. I would submit that such cases should be throughly considered and reviewed.

My point is juniors may take "commandments" too literarily and maybe we should have a disclaimer.

msaspence: I'm thinking mainly of the first two here.

testing-commandments.md
```diff
@@ -0,0 +1,22 @@
...
+If I can delete code and it's corresponding test suite still passes, it's no good.
+
+## Thy broken tests should be helpful
+If I break a feature and a test fails, but it isn't obvious what is wrong, it's no good.
+
+## Thy tests should be consistent
+If a test inconsistently fails, it's no good.
+
+## Thy tests should be fast
+If a unit test takes more than 2 seconds to run, it's no good.
```
msaspence: Whilst there should be no upper limit on the size of a test suite, it might be worth mentioning that test suites that are unnecessarily large incur a cost of maintenance and deployment.

Small test suites (that don't sacrifice too much coverage) have distinct benefits.

testing-commandments.md
theninj4: I'll add a disclaimer :+1:

msaspence:
Are we throwing the baby out with the bath water. Are there things of value in the previous doc that can be lifted, potentially into a separate supporting page.

I love the idea of the commandments but I wonder if we need a supporting document, that is advisory rather than prescriptive for the most part, but does dive into some of the details. I'm thinking specifically as we are all at different stages and junior developers might need a bit more of a steer than this.

Nick-Patrick:
I'd agree with a supporting document. Baring in mind some people join the company who have never touched a test file before. Some recommended practices are still potentially useful.

testing-commandments.md
theninj4: Any big changes where we want to change people's mindset and alter our way of working should be communicated in person to people BEFORE any time is put in to writing code. Using a PR to ultimately change things is inevitable, but it should be the last step in the process. I'll reword the sentence slightly.

testing-commandments.md
theninj4: It's done (see line below)

testing-commandments.md
theninj4: I stand by the clarification of the ambiguity on the line below - `If I make a single code change and many tests fail, it's no good.`. If you can provide a clear definition that doesn't use the words `coupled` or `interface` that forms a better definition, I'm open to hearing it :+1:

testing-commandments.md
theninj4: I'm against using the word `coverage` to influence how tests are written, and suggesting people write the bare minimal set of tests will contradict some of the previous commandments.

msaspence: I agree, but wonder if we should cover the idea of value vs cost of tests

I agree we shouldn't be suggesting people aim for the "bare minimum" but having unnecessarily costly tests that only add diminishing confidence on functionality should be considered.

testing-commandments.md
msaspence: What if, for example, I change the return value of a time orientated module from a timestamped integer to a `Date` object? That is a code change and if that return value is under test, which it should be, my tests will require changing. Which is ok.

I assume by code, you mean code implementation, when the interface could also be interpreted as "code".

theninj4: Yes, your example is Ok. However, if you make that change and 50 tests fail, it's not Ok.

I'm using `Code` to mean the thing that's being tested.

theninj4: I've pushed up a definition on the wording, it's at the top of the doc. I'm open to trying to work it into these points if you think it's worth it.

msaspence: If code is "only the thing being tested" then this is contradictory. IE changes to the thing being tested _should_ break the tests, or at least the one test that tests that thing.

Interface and implementation are terms in common usage in the industry on this subject and I don't think we should avoid using them.

tbh I'm speculating on what people with less experience in testing might need the wording of this to say. I totally agree with the sentiment of the commandment. It would be useful to get wider feedback from other devs on their reading of it.

testing-commandments.md
theninj4: :+1: I've added commandment number 8 below

testing-commandments.md
theninj4: Spence and I have talked over this in person, I've added a small clarification on the end of the line below - `Only test for something once.` I've also reworded `many` to `multiple` to avoid any confusion with `any/many`.

theninj4:
I've restored `testing-standards` and renamed it to `testing-suggestions`

testing-commandments.md
```diff
@@ -0,0 +1,30 @@
...
+`code` is used in reference to the code being tested. The target, maybe?
+`test` is used in reference to the code that is exercising the codebase to validate correctness.
+
+## Thou shalt be consistent within projects
+If a project has a de-facto standard, it should be maintained. The process for changing a de-facto standard should be people focused, and should not be initially communicated via a pull request.
+
+## Thy tests should be obvious
+If the test behaviour isn't obvious by looking at it, it's no good. If I have to scroll for days to read a test, it's no good.
+
+## Thy tests should only check for something once
```
kevinhodges: I don't think it's unreasonable to have 1 line of code covered by more than 1 test, it may be the nature of the branching that follows it that causes this to arise.

theninj4: If I reword it to `Thy tests should only assert for something once` would it be better?

msaspence: `Thee should avoid testing the same thing more than once`?

testing-commandments.md
```diff
@@ -0,0 +1,30 @@
...
+If I make a single code change and multiple tests fail, it's no good.
```
michaelcarter, line 15: I think get what you're trying to say here, but I don't think this sentence does a good job at explaining that. If we were strictly unit testing everything then this would make sense, but with any level of integration testing a single code change may break many tests. especially if it's in a shared module/library reused several times in a larger project.

I interpreted the message behind this as "Tests cost money to write, understand, maintain and run. So don't double up on testing the same thing several times" . Perhaps an alternative description could be _"If a test suite checks the same behaviour several times, it's no good"_?

testing-commandments.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Broken code should break thy tests
+If I can delete code and it's corresponding test suite still passes, it's no good.
```
michaelcarter, line 18: Deleting code and having the corresponding test suite still pass might be an indicator of a great little refactor! This sentence makes more sense to me if you swap `delete` with `break`.

michaelcarter:
A couple of points but otherwise I think this is a great basis for people to start writing better tests from. 👍

testing-commandments.md
theninj4, line 18: I disagree with `Deleting code and having the corresponding test suite still pass might be an indicator of a great little refactor`. If someone can delete code and, without touching the test suite, the tests still pass, that's a red warning light that the tests are no good. It's an indicator that I could go in and make what I believe to be constructive changes to the behaviour of a module, fundamentally break the module, yet get a big green light on the test suite.

Nick-Patrick, line 18: This really depends on what you mean by "delete". The whole point of unit tests is so that you can refactor code safely and hopefully not have to touch the test. If that refactor includes "deleting" code then we don't want the test to suddenly fail.

shackpank, line 18: I'm guessing it was implied but not stated that in this case you would also need to replace the code you deleted with some different code implementing the same behaviour

testing-commandments.md
kevinhodges: agree with you both, something around use case or intended functionality

testing-commandments.md
msaspence, line 18: I would argue that 'deleting' is not the key point here, ANY changes to the code that alter the interface in a way that means it no longer meets the specification is expected should break the tests.

theninj4, line 18: I think we're all agreeing with what is written here? Given a test file `A` and a file full of code `B`, if I go in and delete code from `B`, I expect the test suite to fail. It makes it clear that we want people to write tests to cover their core functionality to prevent other developers from accidentally breaking stuff, it compliments the last commandment (`Thy tests should be described via emergent properties`), and it also leaves the "how" open to creative thinking.

testing-commandments.md
theninj4, line 15: I've tweaked the heading to say `assert` rather than `check`. I'm trying to avoid repeating the heading in the descriptive text. I think using `assert` in the heading should help clarify the body text?

testing-commandments.md
msaspence, line 18: I am certainly an in agreement on the underlying principle. I would contest that the phrasing on this isn't quite right (esp for a junior).

However I don't want this to get stuck in endless debate though so am happy as it is, we can always tweak these in the future.

theninj4:
Ok cool, it feels like we're largely in agreement on the general ideas I'm trying to write down, although there may be a better way of expressing some of them. I'm going to suggest we run some new starters through the list once they're settled in and tweak the definitions based on which ones lead to confusion.

Thanks to everyone for contributing :+1: I'm going to merge this and we can tweak away in other PR's.

PR #131, opened 2016-06-27
Created a tech debt page

msaspence:
Creates a placeholder as an intent to document a strategy for dealing with technical debt.


hx-markterry:
happy with this

PR #133, opened 2016-07-07
Less annoying `consistent-return`

pmcnr-hx:
The base [`consistent-return`](http://eslint.org/docs/rules/consistent-return) rule forced us to add some redundant and unsightly returns (think `return` before `async` as last statement in a function). It turns out that in ESLint 3, this can be configured to be less strict and avoid enforcing those returns that aren't really potential errors and deter from code readability.

A very common pattern for us in the back-end is to callback early in a function in case of error and then end the normal processing with asynchronous code for which we don't care about the return value but which the rule previously forced us to return explicitly anyway.

Previously the rule enforced:

``` javascript
function foo(callback) {
  if (errorCondition) {
    return callback('error');
  }
  return bar(callback);
}
```

When we don't really care about the return value of `bar` and it feels unnatural.

With the amended rule we can do the more natural, readable and explicit:

``` javascript
function foo(callback) {
  if (errorCondition) {
    return void callback('error');
  }
  bar(callback);
}
```

Notice the use of the [`void`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) operator to evaluate the callback but return `undefined` rather than the return value of the callback function and therefore be entirely consistent about the return value of the function (`explicit > implicit`).

Or if preferred:

``` javascript
function foo(callback) {
  if (errorCondition) {
    callback('error');
    return;
  }
  bar(callback);
}
```

Also takes the opportunity to convert to the new (more readable) string based `"off"/"error"` syntax over the old `0/2`.
- [ ] 👍 ShortBreaks
- [x] 👍 Airports


hx-markterry:
ok with me, we'll have to update make-up a bit to jump to eslint v3 as the tests are failing with it currently: https://github.com/holidayextras/make-up/pull/64

hxpaul:
Looks like the "or if preferred" option works with current version of make-up, I'm using it, unless I'm missing something

```
function foo(callback) {
  if (errorCondition) {
    callback('error');
    return;
  }
  bar(callback);
}
```

pmcnr-hx:
The second version works with the current version of `make-up`/`eslint`. You reckon it will be clear? Will developers understand what is going on?

hxpaul:
to me it's clear, maybe open it up further... I'm fine with the change but would probably still rather use `return;` than `return void ...`

pmcnr-hx:
Thanks @hxpaul. I'll create a poll to find out which one is preferred.

PR #134, opened 2016-07-19
Client-side Dependencies

elliotstokes:


connormeredith:
If this relates only to clientside dependencies it might be worth renaming the file to reflect that, or adding it as the heading :)

dancork:
Despite not bundling the JS it could have performance impacts serverside too.

dancork:
Looks good to me, might be worth adding something about choosing modules carefully too. Eg Not bloated with stuff we don't need, regularly maintained, yada yada.

hx-markterry:
The content here is good, but I think we should move it under this heading: https://github.com/holidayextras/culture/blob/master/general-javascript-best-practices.md#dependencies

elliotstokes:
That sounds like a good idea - ill move it now.

hx-markterry:
OK with me

PR #137, opened 2016-07-29
Explanation of Mob Testing procedure

GeorgiaDrew:
# What does this change?

Added explanation of Mob testing procedure and added a link to it on the read me
# Why?

Easier to access for new starters and all in one place


JackMoggy:
Would you say Mob testing could potentially have too many people involved at one time or is there guidelines on that at all do you know?

GeorgiaDrew:
From the conference myself and Andy attended there was a clear message that you can't really have too many people involved in a Mob Test but would need to alter the way you do it. So for example if there is 10 or more of you - you would take the approach of either splitting off into smaller groups or all working together and having one person drive the laptop and rotate between voicing your opinion on what to test/check. Good point to raise though

JackMoggy:
I think this is something we've done before in that we have split into two groups for example with slightly different focuses. Whether or not that is worth adding to the doc? :)

JackMoggy:
👍

rebeccacox:
👍  Looks good to me!

hxmattobee:
I wonder if it's worth highlighting that mob testing isn't an excuse to leave all testing to the last minute. We should still be testing early and often so that we can catch problems when they're still easy and cheap to fix.

hxmattobee:
👍

PR #138, opened 2016-08-05
Added in ES6 classes to working-with-react

owennicol:
This adds in an additional 'ES6 class' version of the **good** React.createClass example.

---
#### Reviewers

**Review 1**
- [ ] :+1:

**Review 2** *
- [ ] :+1:

**Review 3** _(optional)_
- [ ] :+1:

By adding a +1 you are confirming you have...

\*  for HX this review must be completed by an SE, SA or Project Guru

### DIFF
--- a/working-with-react.md
+++ b/working-with-react.md
@@ -110,7 +110,7 @@ The above example is good because the `ContrivedExample` component is no longer 
 
 Using ES6 classes, we can rewrite the above **good** example as follows:
 ```javascript
-class ContrivedExample extends React.Component{
+class ContrivedExample extends React.Component {
 
   handleClick() {
     this.setState({ clicked: true });
@@ -122,7 +122,7 @@ class ContrivedExample extends React.Component{
   …
 };
 
-class TrackableContrivedExample extends React.Component{
+class TrackableContrivedExample extends React.Component {
 
   constructor(){
     // if you use `this` in your method, you have to bind 
--- a/working-with-react.md
+++ b/working-with-react.md
@@ -104,7 +104,43 @@ var TrackableContrivedExample = React.createClass({
   …
 });
 ```
-The above example is good because the `ContrivedExample` component is no longer coupled to anything, it has an optional `onClick` function prop. This makes it suitable for use anywhere. In order to avoid loads of code duplication we've created a second `TrackableContrivedExample` component, which includes the tracker and is suitable for use on projects where we we want a trackable component.
+The above example is good because the `ContrivedExample` component is no longer coupled to anything, it has an optional `onClick` function prop. This makes it suitable for use anywhere. In order to avoid loads of code duplication we've created a second `TrackableContrivedExample` component, which includes the tracker and is suitable for use on projects where we want a trackable component.
+
+**ES6 classes**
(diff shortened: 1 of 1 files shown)


working-with-react.md
```diff
@@ -104,7 +104,43 @@ var TrackableContrivedExample = React.createClass({
...
+    }
+  }
+	
+  …
+};
+
+class TrackableContrivedExample extends React.Component{
+
+  constructor(){
+    // if you use `this` in your method, you have to bind
```
robhuzzey, line 128: Would it be worth suggesting the use of this module to do this automatically: https://www.npmjs.com/package/react-autobind

working-with-react.md
```diff
@@ -104,7 +104,43 @@ var TrackableContrivedExample = React.createClass({
...
 });
 ```
-The above example is good because the `ContrivedExample` component is no longer coupled to anything, it has an optional `onClick` function prop. This makes it suitable for use anywhere. In order to avoid loads of code duplication we've created a second `TrackableContrivedExample` component, which includes the tracker and is suitable for use on projects where we we want a trackable component.
+The above example is good because the `ContrivedExample` component is no longer coupled to anything, it has an optional `onClick` function prop. This makes it suitable for use anywhere. In order to avoid loads of code duplication we've created a second `TrackableContrivedExample` component, which includes the tracker and is suitable for use on projects where we want a trackable component.
+
+**ES6 classes**
+
+Using ES6 classes, we can rewrite the above **good** example as follows:
+```javascript
+class ContrivedExample extends React.Component{
```
robhuzzey: Small point... space before the `{` required here (think it would fail linting in "real life" as well")

working-with-react.md
```diff
@@ -104,7 +104,43 @@ var TrackableContrivedExample = React.createClass({
...
+    this.setState({ clicked: true });
+    if (this.props.handleClick) {
+      this.props.handleClick();
+    }
+  }
+	
+  …
+};
+
+class TrackableContrivedExample extends React.Component{
```
robhuzzey: Same point about linting with `{`

robhuzzey:
Looks good to me +1

working-with-react.md
owennicol, line 128: As this is a basic example of good practice, and not a full solution, I wanted to keep it plugin agnostic and just using React core.

msaspence:
This is going to get brought up at the front end guild meeting on monday. Not trying to block or discourage conversation here but encourage people here to call in for the guild hangout

working-with-react.md
```diff
@@ -104,7 +104,43 @@ var TrackableContrivedExample = React.createClass({
...
+    // it to the method in the constructor
+    // e.g. this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick() {
+    tracker.track('Tracked a click');
+  }
+	
+  render() {
+    return (<ContrivedExample handleClick={() => this.handleClick()} />);
```
djbeaumont, line 138: This might be a bit confusing - in this example is it right that we don't need to bind the method in the constructor since we call it in an arrow function?

PR #141, opened 2016-09-06
Restore block spacing rules

hx-markterry:

This work puts two missing rules back, these rules were removed in error when eslint was upgraded.

See this commit: https://github.com/holidayextras/culture/pull/105/commits/6847b0b871c747f8bfadcb05ce003ac8bd0e44eb for when the problem was introduced.

N/A
#### How can this be tested?

The following code currently passes on the (broken) current linting rules.

``` javascript
'use strict';

var a = function(){
  return 0;
};

var b = function (){
  return 0;
};

var c = function() {
  return 1;
};

a();
b();
c();
```

With this change it will now fail, as only the third example is what we expect.

giphy

---

**Reviewer**
- [x] :+1:
- [x] I don't think this PR needs any additional reviewers

By adding a +1 you are confirming you have...

---

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -141,6 +141,8 @@
     "react/wrap-multilines": 2,
     "semi": [2, "always"],
     "semi-spacing": [2, {"before": false, "after": true}],
+    "space-before-blocks": [2, "always"],
+    "space-before-function-paren": [2, "never"],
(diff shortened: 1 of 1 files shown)


Joezo:
👍

kevinhodges:
👍

PR #143, opened 2016-09-12
Add NPM Command Guidelines

michaelcarter:
This is a writeup after my findings from this doc: https://docs.google.com/spreadsheets/d/1xLrnZWutHbsdHOEN62NKtIggoyhVMd1BTMsa0n5UT_Y/edit#gid=723274934.

These are just my loosely held strong opinions at this stage. All feedback encouraged and welcome!

### DIFF
--- a/npm-command-guidelines.md
+++ b/npm-command-guidelines.md
@@ -1,0 +1,60 @@
+# NPM Command Structure
+
+##What is this?
+This document is an attempt to loosely standardise the names of npm commands we use within our JavaScript projects. This is being proposed to introduce a level of familiarity between projects so that:
+
+ - The learning time common project commands for new developers is shortened.
+ - Context switching between projects is less of a headache.
+ - Developers make fewer mistakes during project setup.
+ - Boilerplate project code becomes more reusable, shortening time to create new projects and encourage similarity between projects at the same time.
+
+This guide has been developed by analysing which npm commands and command patterns are already in use at Holiday Extras, and doing our best to preserve that while establishing a common standard.
+
+##Observations from our repositories
+The following observations have been made from analysing the scripts blocks of all of the `package.json` files within the Holiday Extras Github organisation's projects.
+
+###Types of command
+Our npm commands tend to be structured into 5 main categories:
+
+ - Build related commands
+ - Test related commands
+ - Run related commands
+ - CI related commands
+ - Deploy/Publish related commands
+
+There's also a long tail of miscellaneous commands.
+
+###Command structure
+We seem to almost universally use `:` as a separator to namespace our commands. E.g. `build:js` and `build:css`.
+
+###Commands to run at a certain time
+Excluding NPM supported hooks (`prestop`, `prestart`, etc), we have a not insignificant number of other commands with names that reflect _when_ tasks should be run. This is in contrast to the majority of our commands have names that reflect _what_ the command does.
+
+##Recommendations
+Based on the above, we've come up with the following recommendations for command names.
+
+### Namespace commands
+Commands should be namespaced sensibly in a `:` separated format where possible. Common namespaces that we should aim to share between our projects should include `build`, `test`, `start`, `ci`, `deploy`, `watch`, `debug` and `publish`.
+
+###Reserve the top level of a namespace for a 'do-all' command
+While we do want to namespace, we don't want to stop people from using the top-level `npm run test` or `npm run build` commands. These should, however, be reserved for a 'do everything' task _if_ one is readily available.
+
+For example, `npm run test` should run _all_ the tests of _all_ types, whereas you may have `npm run test:unit`, `npm run test:integration` and `npm run test:journey` if you wanted to split up your test types into individual commands.
+
+###Keep command names technology agnostic
+We shouldn't need to update all of our command names if we decide to switch technologies, and new developers shouldn't have to be familiar with a new technology just to get a project up and running. In support of that, we should use command names that don't reflect the underlying technology where it's not necessary:
+
+**Bad:**
+
+```
+npm run karma
+```
+
+**Good:**
+
+```
+npm run test:browser
+```
+
+###Keep commands reflecting the _what_ as well as _when_ where possible
+Commands names should succinctly reflect what the command is doing, not just when it should be run. Adding a namespace can help here. For example, instead of `npm run postci`, `npm run postci:clear-cache` might be a better command name that gives more insight as to what's going on. The obvious exception to this is [NPM supported hooks](https://docs.npmjs.com/misc/scripts).


npm-command-guidelines.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+**Good:**
+
+```
+npm run test:browser
+```
+
+###Keep commands reflecting the _what_ as well as _when_ where possible
+Commands names should succinctly reflect what the command is doing, not just when it should be run. Adding a namespace can help here. For example, instead of `npm run postci`, `npm run postci:clear-cache` might be a better command name that gives more insight as to what's going on.
```
shackpank: Will that command continue to work as before? Pre and post commands are a hook convention provided by npm. If I've got the [docs](https://docs.npmjs.com/misc/scripts) right, `postci` will run automatically after a `ci` script, but `postci:clear-cache` will run only after a `ci:clear-cache` script, which probably doesn't exist as a `ci` "sub-script". Our projects probably aren't (or probably shouldn't be) running `npm run postci` directly.

npm-command-guidelines.md
```diff
@@ -0,0 +1,60 @@
...
+We seem to almost universally use `:` as a separator to namespace our commands. E.g. `build:js` and `build:css`.
+
+###Commands to run at a certain time
+We have a not insignificant number of commands with names that reflect _when_ they should be run, e.g. `prestart`, `pretest` and `postci`. This is in contrast to the majority of our commands have names that reflect _what_ the command does.
+
+##Recommendations
+Based on the above, we've come up with the following recommendations for command names.
+
+### Namespace commands
+Commands should be namespaced sensibly in a `:` separated format where possible. Common namespaces that we should aim to share between our projects should include `build`, `test`, `run`, `ci`, `deploy`, and `publish`.
```
shackpank: Should `run` here be `start`? (Or we're looking at `npm run run:...`)

theninj4:
Can we get a `tl;dr` at the top for a quick summary? Maybe a copy+paste block for a set of "getting started" npm scripts?

npm-command-guidelines.md
```diff
@@ -0,0 +1,60 @@
...
+###Types of command
+Our npm commands tend to be structured into 5 main categories:
+
+ - Build related commands
+ - Test related commands
+ - Run related commands
+ - CI related commands
+ - Deploy/Publish related commands
+
+There's also a long tail of misellaneous commands.
```
pmcnr-hx: `misellaneous` → `miscellaneous`

npm-command-guidelines.md
```diff
@@ -0,0 +1,60 @@
...
+
+###Command structure
+We seem to almost universally use `:` as a separator to namespace our commands. E.g. `build:js` and `build:css`.
```
pmcnr-hx, line 28: 👍 and is consistent with the `rake` convention for namespacing.

npm-command-guidelines.md
```diff
@@ -0,0 +1,60 @@
...
+
+###Commands to run at a certain time
+We have a not insignificant number of commands with names that reflect _when_ they should be run, e.g. `prestart`, `pretest` and `postci`. This is in contrast to the majority of our commands have names that reflect _what_ the command does.
```
pmcnr-hx: Aren't these hooks?

npm-command-guidelines.md
```diff
@@ -0,0 +1,60 @@
...
+##Recommendations
+Based on the above, we've come up with the following recommendations for command names.
+
+### Namespace commands
+Commands should be namespaced sensibly in a `:` separated format where possible. Common namespaces that we should aim to share between our projects should include `build`, `test`, `run`, `ci`, `deploy`, and `publish`.
+
+###Reserve the top level of a namespace for a 'do-all' command
+While we do want to namespace, we don't want to stop people from using the top-level `npm run test` or `npm run build` commands. These should, however, be reserved for a 'do everything' task _if_ one is readily available.
+
+For example, `npm run test` should run _all_ the tests of _all_ types, whereas you may have `npm run test:unit`, `npm run test:integration` and `npm run test:journey` if you wanted to split up your test types into individual commands.
```
hxpaul, line 42: in SB we agreed to namespace `coverage` as `test:coverage`... SB scripts are here https://github.com/holidayextras/the-works/blob/staging/package.json#L20 - some optimisations still to be made I'm sure. . Some explanation:

```
    "ci": "istanbul cover _mocha -- -R xunit > $CIRCLE_TEST_REPORTS/unit/results.xml", // could be npm run test:coverage -- ...
    "postci": "istanbul check-coverage", // so runs after the ci
    "lint": "make-up lib config controllers flows plugins test",
    "prerelease": "node_modules/deployment-helpers/nodeApps/preRelease.sh",
    "start": "scripts/startup.sh", // this is non standard, because the start command is too long to go here
    "pretest": "npm run prerelease",
    "test": "npm run lint && npm run test:coverage",
    "test:coverage": "istanbul cover _mocha", // works because we have config in .istanbul.yml and mocha.opts
    "posttest": "istanbul check-coverage" // always runs after test
```

can't think of a better name than `prerelease` for that, though it breaks that "don't call pre / post scripts directly" rule. it's possibly not standard enough to worry about though.

then mocha.opts has things that were passed in in the package.json

```
--timeout 15000
--recursive
--reporter spec
```

(plus also `--require test/helpers`) and .istanbul.yml has

```
instrumentation:
  include-all-sources: true
```

so we're reporting on all code

npm-command-guidelines.md
hxpaul: without going crazy expanding the list `watch` and`test:watch` can be handy and now @robhuzzey suggests `debug` too

npm-command-guidelines.md
michaelcarter: @pmcnr-hx I'll update. Feel free to change typos yourself using the edit button. It takes about as long as commenting on it.

pmcnr-hx: I know... I keep forgetting. 😞

npm-command-guidelines.md
michaelcarter: @hxpaul I'm afraid I'm not sure what you're getting at here! 😕  If it's that having a `test:coverage` command doesn't fit in with what I wrote about `npm run test` running all the test commands - Then on the contrary, I think it actually fits quite well as part of `npm run test`!

In my mind, _if_ we're reporting on code coverage then it seems reasonable to assume we have some sort of lower threshold on coverage that we think is too low, and by extension I think we should fail our test run if our coverage threshold isn't met.

With that in mind it seems pretty sensible to me to roll it into the `test:` command namespace. What do you think? Or did I completely misinterpret your comment 😄

hxpaul (commented):
maybe I misunderstood, yes `test:coverage` definitely runs as part of `npm test` - if all the commands are namespaced in this way then we're good...

PR #145, opened 2016-09-25
Updating comms and ownerships notes in expedite

steve-at-hx:
Updating the ownership and communications sections on the expedite process page following the post mortem on the Heroku outage.


carolineerclark:
Hi Steve,

From the expedite today, the bits that are in your PR are the bits that I was missing (but looking for!) - around the email comms in particular. 

I've read through your PR and happy for it to be merged - do you want to press the button so then it's up to date for the rest of the team?

PR #148, opened 2016-10-19
Update general-javascript-best-practices.md

HeinBehrenshx:
Changed const does not work with 'use strict';


SkylerAvery:
I don't think this is right, what's your reasoning for changing it from `const`?

HeinBehrenshx:
Rather look at linting rules in Render

PR #152, opened 2016-12-15
DynamoDB best practices and caveats

pmcnr-hx:
Document about DynamoDB best practices and caveats.


hx-markterry (approved):
Looks good to me

PR #153, opened 2017-01-03
add-gtm-info

hx-damian-casey:
Adds a link to a Google Doc that I created to share knowledge about using workspaces and versions kind of how we do the PR process in github, but on GTM.


DuncanFenning:
do it 👍

hpoom:
A few thoughts. First that doc is only accessible to HX employees but culture repo is public to the world. If the doc would add value to a wider audience then I suggest making it a markdown file and add it to culture repo. If not keep it private, just not sure on best way to link it when culture repo is public.

Also might be worth getting @holidayextras/shortbreaks eyes on it, as they also use GTM. Might eb the doc works group wide, but if not should probably get SB to add to the docs to save any confusion when somebody comes to use it as a guide for working in GTM. My hope is that doc is a generic guide and not HX specific then we can move it from Google Docs to Markdown and will be applicable group wide.

PR #157, opened 2017-03-02
Added a readme document around making a deployment bundle

robhuzzey:
Adds in documentation on how to bundle a project & deploy to S3.

N/A

#### How can this be tested?
Read the docs.

#### Any tech debt?
No.

#### Screenshots / Screencast
N/A

nope.

- I have checked our general [contributing document](https://github.com/holidayextras/culture/blob/master/CONTRIBUTING.md) and the project specific [contributing document](../blob/master/CONTRIBUTING.md) (if present) and I'm happy for this to be reviewed.

By approving a review you are confirming you have...

---

### DIFF
--- a/deployment-bundle-steps.md
+++ b/deployment-bundle-steps.md
@@ -1,0 +1,135 @@
+# How to setup a project with deployment to S3
+
+1. Create S3 bucket for production / staging (your-bucket-name-production & your-bucket-name-staging). (If using cloudfront, see [optional settings](#optional-settings))
+
+2. Log into [circleci.com](https://circleci.com/) and add your project. Set up the following credentials: (see other members of the team for these values).
+  - Environment Variables
+    - GHUSER
+    - GHPASS
+    - GRAPHITE_API_KEY
+    - GRAPHITE_ENDPOINT_PREFIX
+    - CLOUDFRONT_DISTRIBUTION_ID (if using cloudfront for your production bundle)
+  - AWS CodeDeploy
+    - Access Key Id
+    - Secret Access Key
+
+3. Create relevant files for your project:
+  - [circle.yml](#circleyml)
+  - [scripts/deployS3.sh](#deploys3sh)
+  - [istanbul.yml](#istanbulyml)
+
+4. Add [scripts](#npm-scripts) to your package.json
+
+5. Add [devDependencies](#devdependencies) to your package.json
+
+6. Make sure your tests & linting pass. Merge to staging, see it deploy!
+
+---
+
+##### Optional Settings
+
+1a. (optional) Setup cloud front on the production bucket & add the following to the end of your `deployS3.sh` script:
+```bash
+if [ $1 == 'production' ]; then
+  aws cloudfront create-invalidation --distribution-id $CLOUDFRONT_DISTRIBUTION_ID --paths /\*
+fi
+```
+
+---
+
+#### Examples
+
+##### Npm scripts:
+```json
+"scripts": {
+  "build": "webpack --config ./webpack.config.js",
+  "start": "WEBPACK_DEV=true webpack-dev-server --config ./webpack.config.js --content-base ./demo/ --inline --progress",
+  "lint": "make-up src test -i eslint",
+  "ci": "TEST_REPORT_DIRECTORY=${1:-unit}; istanbul cover _mocha --  test/ -R xunit > $CIRCLE_TEST_REPORTS/$TEST_REPORT_DIRECTORY/results.xml",
+  "prerelease": "node_modules/@holidayextras/deployment-helpers/nodeApps/preRelease.sh",
+  "pretest": "npm run prerelease",
+  "test": "npm run lint && npm run test:coverage",
+  "test:watch": "mocha --watch",
+  "test:coverage": "istanbul cover _mocha",
+  "posttest": "istanbul check-coverage",
+  "postinstall": "npm run build"
+},
+```
+
+##### devDependencies:
+```json
+"@holidayextras/deployment-helpers": "*",
+"babel-cli": "^6.10.1",
+"babel-core": "^6.0.20",
+"babel-eslint": "^7.1.1",
+"babel-loader": "^6.0.1",
+"babel-preset-es2015": "^6.0.15",
+"babel-preset-react": "^6.0.15",
+"babel-preset-stage-0": "^6.0.15",
+"istanbul": "v1.1.0-alpha.1",
+"make-up": "^10.0.0",
+"mocha": "^3.2.0",
+"webpack": "^1.12.2",
+"webpack-dev-server": "^1.12.1"
+```
+
+##### istanbul.yml:
+```yml
+check:
+  global:
+    statements: 100
+    lines: 100
+    branches: 100
+    functions: 100
+```
+
+##### deployS3.sh:
+```bash
+#!/bin/bash
+BUCKET=your-bucket-name-$1
+DIR=./dist/
+
+aws s3 sync $DIR s3://$BUCKET --acl public-read --quiet
+```
+
+##### circle.yml
+```yml
+machine:
+  node:
+    version: 6.9.4
+dependencies:
+  pre:
+    - npm install @holidayextras/deployment-helpers
+    - node_modules/@holidayextras/deployment-helpers/nodeApps/preRelease.sh
+    - aws configure set default.region eu-west-1
+    - aws configure set preview.cloudfront true
+test:
+  pre:
+    - mkdir -p $CIRCLE_TEST_REPORTS/unit
+    - mkdir -p $CIRCLE_TEST_REPORTS/coverage
+  override:
+    - npm run lint
+    - npm run ci
+  post:
+    - npm install coverage-percentage
+    - cp -r ./coverage/ $CIRCLE_ARTIFACTS/coverage
+    - echo "$GRAPHITE_API_KEY.counters.$CIRCLE_PROJECT_REPONAME.production.test.coverage `./node_modules/.bin/coverage-percentage ./coverage/lcov.info --lcov`" | nc -uw0 carbon.hostedgraphite.com 2003
+deployment:
+  production:
+    branch: master
+    commands:
+      - npm run build
+      - ./scripts/deployS3.sh production
+      - node_modules/@holidayextras/deployment-helpers/nodeApps/postRelease.sh production
+  staging:
+    branch: staging
+    commands:
+      - npm run build
+      - ./scripts/deployS3.sh staging
+experimental:
+  notify:
+    branches:
+      only:
+        - master
+        - staging
+```
--- a/deployment-bundle-steps.md
+++ b/deployment-bundle-steps.md
@@ -11,7 +11,7 @@ This document assumes you have already setup the S3 bucket.
     - GRAPHITE_API_KEY
     - GRAPHITE_ENDPOINT_PREFIX
     - CLOUDFRONT_DISTRIBUTION_ID (if using cloudfront for your production bundle)
-  - AWS CodeDeploy
+  - AWS Permissions
     - Access Key Id
     - Secret Access Key
 
--- a/deployment-bundle-steps.md
+++ b/deployment-bundle-steps.md
@@ -1,6 +1,6 @@
 # How to setup a project with deployment to S3
 
-1. Create S3 bucket for production / staging (your-bucket-name-production & your-bucket-name-staging). (If using cloudfront, see [optional settings](#optional-settings))
+This document assumes you have already setup the S3 bucket.
 
 ## Using [CircleCi](https://circleci.com/)
 
--- a/deployment-bundle-steps.md
+++ b/deployment-bundle-steps.md
@@ -2,7 +2,9 @@
 
 1. Create S3 bucket for production / staging (your-bucket-name-production & your-bucket-name-staging). (If using cloudfront, see [optional settings](#optional-settings))
 
-2. Log into [circleci.com](https://circleci.com/) and add your project. Set up the following credentials: (see other members of the team for these values).
+## Using [CircleCi](https://circleci.com/)
+
+1. Log into [circleci.com](https://circleci.com/) and add your project. Set up the following credentials: (see other members of the team for these values).
   - Environment Variables
     - GHUSER
     - GHPASS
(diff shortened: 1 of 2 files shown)


deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+```yml
+check:
+  global:
+    statements: 100
+    lines: 100
+    branches: 100
+    functions: 100
+```
+
+##### deployS3.sh:
```
elliottcrush: Could we also add this script to deployment helpers? I think we should introduce the correct setting of cache headers as well here if we're encouraging a standard.

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+```bash
+#!/bin/bash
+BUCKET=your-bucket-name-$1
+DIR=./dist/
+
+aws s3 sync $DIR s3://$BUCKET --acl public-read --quiet
```
elliottcrush: Potentially add the CF invalidation as a step here as well? I think its unlikely you want to sync to S3 without wanting to clear cloudfront during a deployment.

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+
+5. Add [devDependencies](#devdependencies) to your package.json
+
+6. Make sure your tests & linting pass. Merge to staging, see it deploy!
+
+---
+
+##### Optional Settings
+
+1a. (optional) Setup cloud front on the production bucket & add the following to the end of your `deployS3.sh` script:
```
elliottcrush: These initial setup practices differ between SB and Airports. Perhaps we should just have one step for getting an IAM setup that has access to S3 and CF - then we can add steps for creating buckets and CF setup all through the CLI.....Just a thought

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+    - cp -r ./coverage/ $CIRCLE_ARTIFACTS/coverage
+    - echo "$GRAPHITE_API_KEY.counters.$CIRCLE_PROJECT_REPONAME.production.test.coverage `./node_modules/.bin/coverage-percentage ./coverage/lcov.info --lcov`" | nc -uw0 carbon.hostedgraphite.com 2003
+deployment:
+  production:
+    branch: master
+    commands:
+      - npm run build
+      - ./scripts/deployS3.sh production
+      - node_modules/@holidayextras/deployment-helpers/nodeApps/postRelease.sh production
+  staging:
```
elliottcrush: Do we need staging environments for static applications?

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+6. Make sure your tests & linting pass. Merge to staging, see it deploy!
+
+---
+
+##### Optional Settings
+
+1a. (optional) Setup cloud front on the production bucket & add the following to the end of your `deployS3.sh` script:
+```bash
+if [ $1 == 'production' ]; then
+  aws cloudfront create-invalidation --distribution-id $CLOUDFRONT_DISTRIBUTION_ID --paths /\*
```
shackpank: It's generally going to be preferable to deploy a uniquely named asset than to re-use an existing filename and issue a cache invalidation to it

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+DIR=./dist/
+
+aws s3 sync $DIR s3://$BUCKET --acl public-read --quiet
+```
+
+##### circle.yml
+```yml
+machine:
+  node:
+    version: 6.9.4
```
shackpank: Is it possible to specify just 6, so things like security patches are automatically updated to?

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+  pre:
+    - mkdir -p $CIRCLE_TEST_REPORTS/unit
+    - mkdir -p $CIRCLE_TEST_REPORTS/coverage
+  override:
+    - npm run lint
+    - npm run ci
+  post:
+    - npm install coverage-percentage
+    - cp -r ./coverage/ $CIRCLE_ARTIFACTS/coverage
+    - echo "$GRAPHITE_API_KEY.counters.$CIRCLE_PROJECT_REPONAME.production.test.coverage `./node_modules/.bin/coverage-percentage ./coverage/lcov.info --lcov`" | nc -uw0 carbon.hostedgraphite.com 2003
```
shackpank: Generating coverage and pushing it into graphite like this seems specific to SB - distribution projects that track coverage over time (eg trip-planner) use coveralls

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+"mocha": "^3.2.0",
+"webpack": "^1.12.2",
+"webpack-dev-server": "^1.12.1"
+```
+
+##### istanbul.yml:
+```yml
+check:
+  global:
+    statements: 100
```
shackpank: So your build will fail on 99.9% coverage?

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+---
+
+#### Examples
+
+##### Npm scripts:
+```json
+"scripts": {
+  "build": "webpack --config ./webpack.config.js",
```
shackpank: How "pick up and go" do we expect this to be? This expects that you've done the legwork to create a webpack config, but don't know what dependencies would need to go in `package.json`

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+  "start": "WEBPACK_DEV=true webpack-dev-server --config ./webpack.config.js --content-base ./demo/ --inline --progress",
+  "lint": "make-up src test -i eslint",
+  "ci": "TEST_REPORT_DIRECTORY=${1:-unit}; istanbul cover _mocha --  test/ -R xunit > $CIRCLE_TEST_REPORTS/$TEST_REPORT_DIRECTORY/results.xml",
```
shackpank: This looks like quite a nice circle feature! (collecting metrics provided you drop them into a conventional place)

shackpank (commented):
This feels like a bit of a mixture of things - "how do we give CI the permissions it needs to deploy using our shared scripts, and what S3/env-var naming conventions do we use" strikes me as the more important part that's specific to us; whereas "what should you drop into an empty directory to turn an empty project into something that packages up a js file" is better left to documentation maintained by others or to some "living examples" maintained internally, otherwise this could rapidly become outdated and/or overspecified

deployment-bundle-steps.md
hpoom: This is SB specific and is one of the hard parts with the Culture repo, as there is much Airports specific stuff that has been added to docs that has to be taught into SB engineers to ignore. Ideally we would have the same approach and maybe in time we will, but in the meantime how should we write culture repo docs so they are valuable to all parts of the group. I have some ideas but keen to hear from others too.

deployment-bundle-steps.md
hpoom: See my comment below about SB and Airports differences and how we solve that with culture repo documents. We need to find a solution.

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
+# How to setup a project with deployment to S3
+
+1. Create S3 bucket for production / staging (your-bucket-name-production & your-bucket-name-staging). (If using cloudfront, see [optional settings](#optional-settings))
+
+2. Log into [circleci.com](https://circleci.com/) and add your project. Set up the following credentials: (see other members of the team for these values).
+  - Environment Variables
+    - GHUSER
+    - GHPASS
```
hxpaul: is there an alternative to this one? I remember a comment here https://github.com/holidayextras/deployment-helpers/pull/10 saying "stop using GHPASS"

deployment-bundle-steps.md
hxpaul: `-config ./webpack.config.js` is the default for webpack isn't it? So not required afaik, see for example https://github.com/holidayextras/tracker/blob/staging/package.json

deployment-bundle-steps.md
hxpaul: it is, what's the advantage of assigning TEST_REPORT_DIRECTORY like that? We have been doing
```
"ci": "npm run lint && MOCHA_FILE=$CIRCLE_TEST_REPORTS/junit/test-results.xml istanbul cover _mocha -- -R mocha-junit-reporter && istanbul check-coverage"
```
here for example https://github.com/holidayextras/tracker/blob/staging/package.json#L16

also ci should run the linting shouldn't it?

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+#### Examples
+
+##### Npm scripts:
+```json
+"scripts": {
+  "build": "webpack --config ./webpack.config.js",
+  "start": "WEBPACK_DEV=true webpack-dev-server --config ./webpack.config.js --content-base ./demo/ --inline --progress",
+  "lint": "make-up src test -i eslint",
+  "ci": "TEST_REPORT_DIRECTORY=${1:-unit}; istanbul cover _mocha --  test/ -R xunit > $CIRCLE_TEST_REPORTS/$TEST_REPORT_DIRECTORY/results.xml",
+  "prerelease": "node_modules/@holidayextras/deployment-helpers/nodeApps/preRelease.sh",
```
hxpaul: worth mentioning what preRelease does here as I don't think it's been used in core before. Automating the version updating might be better than manually updating the version number in package.json and fixing conflicts

deployment-bundle-steps.md
hxpaul: I'm all over this one! Have set this barrier on new things we've worked on including ccproxy and provider-tokenise...

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+dependencies:
+  pre:
+    - npm install @holidayextras/deployment-helpers
+    - node_modules/@holidayextras/deployment-helpers/nodeApps/preRelease.sh
+    - aws configure set default.region eu-west-1
+    - aws configure set preview.cloudfront true
+test:
+  pre:
+    - mkdir -p $CIRCLE_TEST_REPORTS/unit
+    - mkdir -p $CIRCLE_TEST_REPORTS/coverage
```
hxpaul: I don't think these are required won't they be created on the fly if they do not exist?

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+  override:
+    - npm run lint
+    - npm run ci
```
hxpaul: add run lint into `run ci` instead

deployment-bundle-steps.md
hxpaul: found the same thing on https://github.com/holidayextras/tracker - it builds a staging version, but nothing is using it

deployment-bundle-steps.md
hxpaul: also you'd probably want to build a minified and non minified version of the code, https://github.com/holidayextras/tracker is a good example I think. Also I think we can use one standard webpack.config.js file mostly shared across repos. Things like the name and entry point of the repo can be passed in as npm vars

deployment-bundle-steps.md
hx-markterry: Yes you need to use token instead of password, if the account has 2FA.

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+##### circle.yml
+```yml
+machine:
+  node:
+    version: 6.9.4
+dependencies:
+  pre:
+    - npm install @holidayextras/deployment-helpers
```
hx-markterry: Can't we just add this to the project's dev dependancies?

deployment-bundle-steps.md
```diff
@@ -0,0 +1,135 @@
+# How to setup a project with deployment to S3
+
+1. Create S3 bucket for production / staging (your-bucket-name-production & your-bucket-name-staging). (If using cloudfront, see [optional settings](#optional-settings))
```
hx-markterry: Could this be more strict about naming here?

In Dist. we have so many buckets we don't know what half of them are for, if you used `repo_name-staging` or `repo_name-production` then it is obvious what is using the bucket.

hxpaul:
It's getting a bit bike shed. I recommend going back to Ollie's point, keeping this doc sharp and to the point, remove everything not strictly related to the original idea. Make it only a list of things that are strictly required to getting a bundle deployed to S3, and remove all the contentious things that are different in different parts of the business. No need to talk about testing or linting here or even webpack probably.

robhuzzey:
@hxpaul sensible point... I hadn't commented yet for this very reason... essentially I gave enough information for the point of how to deploy a bundle, but then the additional info around specifics of how to get to that bundle were what were targeted... bikeshedded. I will likely close this PR & create a new one. Thanks.

deployment-bundle-steps.md
robhuzzey: Cloudfront is optional, not all deployments would use it.

deployment-bundle-steps.md
robhuzzey: This is an optional thing.... the circle.yml example I am using in this doc, is an example.

deployment-bundle-steps.md
robhuzzey: Not necessarily, at times we would need an exact filename like in our case of providing a 3rd party a booking engine like `http://ourdomainfors3/engine.js`

For convenience, we wouldn't want that 3rd party to have to update their url everytime we made a change.

deployment-bundle-steps.md
robhuzzey: This is an example of scripts you can use, and not necessarily used to "pick up & go"... the main points I wanted to get across were how to deploy a bundle.

deployment-bundle-steps.md
robhuzzey: These are example scripts and not strictly related to the bundling / deploy of the app... but... that said, it's what all SB projects use & yes you are right, it seems we are trying to lint twice as it runs `npm run lint` then `npm run ci` in the circle.yml... so good spot! thanks :)

deployment-bundle-steps.md
robhuzzey: It's an example... but yeah... we do try to hit 100%... it can be adjusted on a per project basis.

deployment-bundle-steps.md
robhuzzey: This is not something I'd expect a developer to blindly pick up & use... `6.9.4` is the value as an example here.

deployment-bundle-steps.md
robhuzzey: Additionally, this is an example & I wouldn't want developers to blindly accept examples given here. These values can be tweaked on a per-project basis.

deployment-bundle-steps.md
robhuzzey: I think we need standardisation regarding this across the business as in order to share things, we'd need accepted Environment Variable names. @hpoom this is one place where I do think we need a discussion.

deployment-bundle-steps.md
robhuzzey: I'm not sure how I feel about this... how do we enforce semver... would we automatically update as a patch unless told otherwise? What happens if dev's become dependent on the automation? Do we end up having breaking changes being deployed under a patch revision?

deployment-bundle-steps.md
hxpaul: yep in practice we are doing it as @robhuzzey describes (I know tracker does and the last booking engines I worked on did too) but we probably should be doing it as @shackpank describes for better cache management. For an ultimate-shared-tech-document maybe we should cover both cases or neither. we could upload as both `engine.min.js` and `engine.v3.2.1.min.js` and only bust the cache on `engine.min.js` maybe?

deployment-bundle-steps.md
robhuzzey: I'm not sure... they can be removed on a pre-project basis.

deployment-bundle-steps.md
robhuzzey: Agreed.

deployment-bundle-steps.md
robhuzzey: I recall this was specifically required here... @notABluesSinger can confirm, but I believe the CI needed this explicitly?

deployment-bundle-steps.md
robhuzzey: That's a good point... but I think the project may have differing needs, for example, many projects could share one bucket... how would we deal with that?

deployment-bundle-steps.md
hxpaul: I think this is out of date info, I have seen `npm install -g` in here but just putting it in dev dependencies does the job better and cleaner I think

deployment-bundle-steps.md
hxpaul: no I think Infrastructure have a rule about this so check with @matt-smith for example and then you can tighten this up, probably automate it.

deployment-bundle-steps.md
hxpaul: no just delete it, it's not required in any project afaik, it will be created on the fly I'm sure

deployment-bundle-steps.md
hxpaul: this is just another distraction, it's not relevant to building a bundle really.

deployment-bundle-steps.md
notABluesSinger: Happy to push this data out to coveralls, especially if we're already paying for it as a group.

deployment-bundle-steps.md
notABluesSinger: It is a token, it was just named wrong up front... we can absolutely add the ability to use `GHTOKEN` and fallback to `GHPASS` if needed, that we we can move across.
This variable is consumed by deployment-helpers, so we'd need to adjust that and then update the std here. 
We should absolutely merge these back into one file and one process, shame we ended up with two :/

deployment-bundle-steps.md
notABluesSinger: The reason this is not all in one command is the way that circle breaks them down as tasks... provides a much nicer interface when viewed and you can quickly see what task it failed on
"Failed on `npm run lint`" get's you right to the issue.

deployment-bundle-steps.md
notABluesSinger: We're unable to run the version bump check prior to the `npm install` task without the manual install. 
The reason for wanting to check the version bump before running the `npm install` is to fail fast.

deployment-bundle-steps.md
robhuzzey: @pauly that PR wasn't merged... just looked at the current code for deployment-helpers & it is GHUSER & GHPASS : https://github.com/holidayextras/deployment-helpers/blob/master/nodeApps/postRelease.sh#L10

deployment-bundle-steps.md
```diff
@@ -0,0 +1,39 @@
+# How to setup a project with deployment to S3
+
+This document assumes you have already setup the S3 bucket.
+
+## Using [CircleCi](https://circleci.com/)
+
+1. Log into [circleci.com](https://circleci.com/) and add your project. Set up the following credentials: (see other members of the team for these values).
+  - Environment Variables
+    - GHUSER
+    - GHPASS
```
hxpaul, line 10: nice that you got rid of all the other controversial bits, I would still like someone who has more experience than me with this to confirm that this is the best thing to do, to put GHPASS as an env var in circle.

robhuzzey: @notABluesSinger I think you can answer this?

notABluesSinger: Until we get something in place to store these kind of credentials, such as vault etc, this is the best choice, for me, to store it. Obviously, we need to keep these creds out of VCS, but happy for other suggestions.

hxpaul: I think you have removed the reference to things that might want to use the password, so could safely remove this from this guide can't you? I might be misunderstanding but it looks to me like it says "this is our best practice, don't use two factor authentication here"

robhuzzey:
@hxpaul @hpoom @shackpank @hx-markterry @notABluesSinger Can we merge this now the changes have been made... I'm keen to not let it go stale... thanks all :)

robhuzzey:
@hxpaul @hpoom @shackpank @hx-markterry @notABluesSinger I am going to merge this Monday unless there are any objections.... thanks :)

deployment-bundle-steps.md
notABluesSinger: It's not a password, it's still a token... so the user does have two-factor auth setup. Perhaps a wrongly named variable...

deployment-bundle-steps.md
```diff
@@ -0,0 +1,39 @@
...
+    - GRAPHITE_API_KEY
+    - GRAPHITE_ENDPOINT_PREFIX
+    - CLOUDFRONT_DISTRIBUTION_ID (if using cloudfront for your production bundle)
+  - AWS CodeDeploy
```
notABluesSinger: Isn't this `AWS Permissions`?

hx-markterry (approved):
Happy with this

deployment-bundle-steps.md
robhuzzey: Yes it is! Good spot... I've changed.

PR #164, opened 2017-05-17
working-with-react tiny markdown header fix.

misterdai:


hpoom:
Thanks for the contribution and the fix.

PR #169, opened 2017-10-03
TPM guidance refinement

steve-at-hx:
Following a discussion between the directors and a few tech leads across the team, this is a refinement of the TPM guidance to put more responsibility on the hosts to provide solutions to ideas in advance to be refined and validated in the meeting. Also includes a link to a cut down, easier to circulate version of the meeting template, also for review.

### DIFF
--- a/technical-planning-meeting.md
+++ b/technical-planning-meeting.md
@@ -33,18 +33,18 @@
 
 ## How does the process work?
 
-##### Part 1 - Before the TPM
+### Part 1 - Before the TPM
 
 * **Important update:** Prepare a solution to the problem. Try not to take an open or unconsidered problem case to the group, using it to refine and validate instead.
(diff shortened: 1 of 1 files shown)


nickloestartup:
@steve-at-hx / @hpoom 
The Agile Coaches have reviewed and are happy with this approach. There has been feedback that the process was too complicated to follow and this iteration trims things down nicely.

The only comment is that the business impact section was specifically asked for by Pod Leads. As a result I've asked (via slack channel #pl-channel) that someone comment on behalf of their discipline. I appreciate the guidance indicates this should happen, but removing the section on the document might mean people have the excuse to skip it. Let's hope not.

But in principle, all looks good to us - and some are test-running it already for new TPMs, so can feedback on how they've got on.

hpoom:
Cool once we have feedback and the business impact section resolved we should merge.

gozzatron:
Much prefer the one page layout of the document. 

Seems much more streamlined, and makes sense to take out the agenda prompts (i.e. 5 mins for the summary, 20 mins discussion, etc) as they could/should go into the TPM calendar invite itself. 

For those not too familiar with the task, could we include a suggested length of time for the TPM so that we ensure enough time is put aside to discussing the solutions (i.e. S/M issue, from t-shirt sizing, 30 mins, L 1hr, XL 90mins or question whether we break this up).

PR #171, opened 2018-05-11
:thinking: improve expedite email suggestions

hxpaul:


AmyUttley (commented):
Perfect. Although ¿embiggen? 

How about: 'Use plain English and non-technical words where possible take make it easier for the reader to understand the facts'

gozzatron:
Unsure whether some support content on how to complete each section would help. Simon commented that there have been a number of expedite emails where the summary was that customers were unable to make bookings for 30 minutes - this doesn't cover what actually happened to customers (unable to search, and unable to book are two separate things).

hxpaul:
Embiggen is a perfectly cromulent word https://www.google.co.uk/search?q=embiggen - the use of technical words is fine IMO, we are a tech business. I was joking a bit with "embiggen", it did originate in a Simpsons episode but it is in the dictionary now and the definition covers the thing that has been happening that we should try to avoid.

PR #172, opened 2018-07-23
Formatting, schema.org property names and clarify .NET usage.

pmcnr-hx:
Add clarification to the API task that solutions should have cross platform support, especially in the case of .NET.

Takes the opportunity to tidy up formatting a bit and update the attribute names to schema.org taxonomy with is a more I18N friendly (people's first name aren't necessarily their given name).


djbeaumont (approved):
:shipit:

PR #176, opened 2019-01-22
Clarifications for how the API task needs to be easily consumable.

pmcnr-hx:
Clarifications to the API task, making clear that the solution must be easily consumable by a plain HTTP client or it must be trivial to write a client if using a transport other than HTTP.


jamesmcneil (approved):
Agree with all these points 👍

PR #177, opened 2019-02-05
Remove tpm template

Smuger07:
This template is no longer needed as we have a new one listed here:
https://github.com/holidayextras/culture/blob/master/technical-planning-meeting.md#part-1---before-the-tpm


jamesmcneil (approved):
🗑

PR #185, opened 2019-09-19
Delete application-tracking.md

lukehansell:
removing as out of date with current best practice


lukehansell:
confirmed with Mark Terry that this should be removed

PR #191, opened 2019-11-28
Removing git flow

lukehansell:
we do not support gitflow at HX so this document is unnecessary.
I'm therefore removing it and the references to it as a part of our drive to clean up our culture repo.


hx-elliot-massey (approved):
This change pleases me. 👏

PR #195, opened 2019-12-06
Templates

lukehansell:
moving all of the templates to their own folder for now. Some of these are still referenced. I've made updates to a few of them to make sure they're up to date with our practices.


cameronviner (approved):
Looks good 👍

PR #196, opened 2019-12-06
Tech debt

lukehansell:
moving tech debt into best practices and correcting the links from the best-practices README


cameronviner (approved):
Looks good 👍

PR #200, opened 2025-08-06
Update technical-resources.md - remove outdated resources

harrygates23:
This is a short term step to avoid suggesting use of tools like moment and lodash, which we just dont want anymore.


technical-resources.md
```diff
@@ -36,8 +22,6 @@ The links below should provide a good introduction into an item, further in dept
 ### [HapiJS](http://hapijs.com/)
```
richardwrighthx: should this be removed as well?

harrygates23: good spot. ive made some more changes 👍

